PR #1, opened 2016-08-08
DO NOT MERGE: this is where you add groups

deads2k:
@gabemontero This is how the groups can be wired in and I added a comment that describes how to sanely ACL access to jenkins.

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -42,6 +42,7 @@ import org.acegisecurity.AuthenticationException;
 import org.acegisecurity.AuthenticationManager;
 import org.acegisecurity.BadCredentialsException;
 import org.acegisecurity.GrantedAuthority;
+import org.acegisecurity.GrantedAuthorityImpl;
 import org.acegisecurity.context.SecurityContextHolder;
 import org.acegisecurity.providers.UsernamePasswordAuthenticationToken;
 import org.acegisecurity.providers.anonymous.AnonymousAuthenticationToken;
@@ -257,11 +258,13 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
                                 }
                             });
                     GenericUrl url = new GenericUrl(serverPrefix + "/oapi/v1/users/~");
-
+                    
+                    // do self-sar here to see if you have the "admin" or "view" verb in the "build.openshift.io" api group on the "jenkins" resource 
+                    
                     HttpRequest request = requestFactory.buildGetRequest(url);
 
                     OpenShiftUserInfo info = request.execute().parseAs(OpenShiftUserInfo.class);
-                    GrantedAuthority[] authorities = new GrantedAuthority[]{SecurityRealm.AUTHENTICATED_AUTHORITY};
+                    GrantedAuthority[] authorities = new GrantedAuthority[]{SecurityRealm.AUTHENTICATED_AUTHORITY, new GrantedAuthorityImpl("edit")};
                     // logs this user in.
                     UsernamePasswordAuthenticationToken token =
                             new UsernamePasswordAuthenticationToken(info.getName(), "", authorities);


deads2k:
Along with https://github.com/openshift/jenkins/pull/133, this would allow someone with the correct access to have view authority in jenkins.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -257,11 +258,13 @@ public void initialize(HttpRequest request) throws IOException {
                                 }
                             });
                     GenericUrl url = new GenericUrl(serverPrefix + "/oapi/v1/users/~");
-
+                    
+                    // do self-sar here to see if you have the "admin" or "view" verb in the "build.openshift.io" api group on the "jenkins" resource
```
gabemontero: @deads2k since I didn't find references to the string "build.openshift.io" in the origin repo, I'm assuming that api group is not available yet.  Is that correct?

Would the url start with `/oapi/build.openshift.io/v1` ? .... analogous to `/apis/extension/v1` perhaps ...

deads2k: You'll hit the `oapi/v1/subjectaccessreviews` with content like

```json
{"kind":"SubjectAccessReview","apiVersion":"v1","namespace":"YOUR_NAMESPACE","verb":"admin","resourceAPIGroup":"build.openshift.io","resourceAPIVersion":"","resource":"jenkins","resourceName":"","content":null,"user":"","groups":[],"scopes":null}
```

gabemontero: tried a quick experiment, that failed:
```
gmontero ~ $ oc rsh jenkins-1-xm9w1
sh-4.2$ oc get subjectaccessreviews
oc get subjectaccessreviews
Error from server: User "system:serviceaccount:test:default" cannot list subjectaccessreviews in project "test"
sh-4.2$ 
```

Invalid experiment?

deads2k: It's a POST.  `oc policy can-i`

gabemontero: Sure, I assumed the JSON from your prior comment was a POST body, but you can't GET things to see what has been previously posted?  Or will the GET only work after something has been POSTed ?

gabemontero: Also, as it happens the `oc` in the jenkins image is old enough that it does not have `oc policy can-i`.

@bparees FYI - it may be time to update the `oc` binary in the Jenkins image.

bparees: v1.2.1 is the latest release we've got:
https://github.com/openshift/origin/releases/tag/v1.2.1
and that's what the image is using:
https://github.com/openshift/jenkins/blob/master/1/Dockerfile#L36

bparees: (ignoring the 1.3.x alpha releases)

gabemontero: So I can't figure out how to get permssion to create the subjectaccessreview ... with the url ending in `oapi/v1/subjectaccessreviews` I get a message saying I can't create those at the cluster scope.  I tried inserting a `namespaces/<my namespace>` in the appropriate spot, and it says I can't create a subjectaccessreview in namespace specified by `<my namspace>`

What roles do I need to add to the user I'm logging in as?

deads2k: > So I can't figure out how to get permssion to create the subjectaccessreview ... with the url ending in oapi/v1/subjectaccessreviews I get a message saying I can't create those at the cluster scope. I tried inserting a namespaces/<my namespace> in the appropriate spot, and it says I can't create a subjectaccessreview in namespace specified by <my namspace>
> 
> What roles do I need to add to the user I'm logging in as?

This request should be made using the token you've just gotten for the user.  

What does your payload look like?  The username and groupname should be empty.  All authenticated users are allowed to create subjectaccessreviews at the cluster scope as long as the username and groupname are empty

gabemontero: Yep, I'm using the token that was received.  Here is some debug I added which prints the code from the resposne, and then the access token after it is passed into the `Credential` class:

```
Aug 17, 2016 6:08:45 PM FINER org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm doFinishLogin
ENTRY code=IDrXlGRzCSPpapR_KkJjsl3Zupqd6zsqJD8aT28XX2o&state=ZjVhOTNkMGEtOTUyMS00 http://172.30.123.81:8080/securityRealm/finishLogin
Aug 17, 2016 6:08:45 PM FINE org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm
GGM auth code provided IDrXlGRzCSPpapR_KkJjsl3Zupqd6zsqJD8aT28XX2o
Aug 17, 2016 6:08:45 PM FINE org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm
GGM credential's access token is j8WhkfpoFSzVPOAifIQumSfIBZaURSUbkZFZueg9A2I
```

Here is the json I took from your earlier response and converted to a Java string that is leveraged by the `HttpContent` body for the POST:

```
    	final String json = "{\"kind\":\"SubjectAccessReview\",\"apiVersion\":\"v1\",\"namespace\":\""+namespace+"\",\"verb\":\"admin\",\"resourceAPIGroup\":\"build.openshift.io\",\"resourceAPIVersion\":\"\",\"resource\":\"jenkins\",\"resourceName\":\"\",\"content\":null,\"user\":\"\",\"groups\":[],\"scopes\":null}";
```

I'm running with  `AllowAllPasswordIdentityProvider`

I logged in via oc as `test` to create the project / namespace that Jenkins is running in and that I am creating buildconfig's, deploymentconfigs's, etc.

I then log into Jenkins successfully as test via the OpenShift login, and have the create subjectaccessreview attempt surrrounded by a `try { ... } catch (Throwable t) { .. }`

Here is the precise `com.google.api.client.http.HttpResponseException: 403 Forbidden` failure message:  

```
com.google.api.client.http.HttpResponseException: 403 Forbidden
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "scopes [user:info] prevent this action; User \"test\" cannot create subjectaccessreviews at the cluster scope",
  "reason": "Forbidden",
  "details": {
    "kind": "subjectaccessreviews"
  },
  "code": 403
}
```

deads2k: > "message": "scopes [user:info] prevent this action; User \"test\" cannot create subjectaccessreviews at the cluster scope",

Yay authorization! :)

The scopes on your token aren't sufficient.  You need `user:check-access` too

gabemontero: OK progress ... after searching through doc/code, I found the `Scopes` fields in three token types in `pkg/oauth/api/types.go`.  I found an `Add` method in `pkg/oauth/scope/scope.go`, which is used by the `GrantHandler` interface.  I've only found HTTP POST level access to the `GrantHandler`, not a direct `oc` or `oadm` entryway into adding scopes (unlike roles).

So I ran an `oc get` for the 3 oauth types that had scope fields ... two get hits:
```
gmontero ~ $ oc get oauthaccesstoken -o json
{
    "kind": "List",
    "apiVersion": "v1",
    "metadata": {},
    "items": [
        {
            "kind": "OAuthAccessToken",
            "apiVersion": "v1",
            "metadata": {
                "name": "txcQaEHfabHd2N4-zNEcdALalTkoHSDHRCxXKI05Fzo",
                "selfLink": "/oapi/v1/oauthaccesstokens/txcQaEHfabHd2N4-zNEcdALalTkoHSDHRCxXKI05Fzo",
                "uid": "6635559f-6556-11e6-a4ef-68f7287398a7",
                "resourceVersion": "499",
                "creationTimestamp": "2016-08-18T15:14:03Z"
            },
            "clientName": "openshift-challenging-client",
            "expiresIn": 86400,
            "scopes": [
                "user:full"
            ],
            "redirectURI": "https://192.168.0.9:8443/oauth/token/implicit",
            "userName": "test",
            "userUID": "6634711e-6556-11e6-a4ef-68f7287398a7"
        }
    ]
}
gmontero ~ $ oc get oauthclientauthorization -o json
{
    "kind": "List",
    "apiVersion": "v1",
    "metadata": {},
    "items": [
        {
            "kind": "OAuthClientAuthorization",
            "apiVersion": "v1",
            "metadata": {
                "name": "test:system:serviceaccount:test:default",
                "selfLink": "/oapi/v1/oauthclientauthorizations/test:system:serviceaccount:test:default",
                "uid": "53d50d3a-655f-11e6-a4ef-68f7287398a7",
                "resourceVersion": "1382",
                "creationTimestamp": "2016-08-18T16:17:58Z"
            },
            "clientName": "system:serviceaccount:test:default",
            "userName": "test",
            "userUID": "6634711e-6556-11e6-a4ef-68f7287398a7",
            "scopes": [
                "user:info"
            ]
        }
    ]
}
```
One is `user:full` and just seems to reference the `test` id, the other seems to be for the default serviceaccount but also references the `test` id and is `user:info`

I re-ran my test, but the token codes returned don't match the token for the default service account.  Should they match up?

I'll circle back and see about updating the client auth for the default sa I found with user:info, but wanted to at least confirm if I'm on the right path and if you can say from what I provided that things are working correctly?

thanks

deads2k: I think you need to get an additional scope specified here: https://github.com/deads2k/jenkins-openshift-login-plugin/blob/e5449846231b32688c23e904593fbf374c82d13e/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java#L90

gabemontero: OK that helped - thanks.  Inching closer ... seems like I'm close.  Adding both scopes (had to otherwise the get of the `User` failed without `user:info`) I new get a response from the POST of subjectaccessreview instead of the Java access exception, but it still failed.

The http response was:
```
GGM response from post {"kind":"SubjectAccessReviewResponse","apiVersion":"v1","namespace":"test","allowed":false,"reason":"scopes [user:check-access user:info] prevent this action; User \"test\" cannot \"admin\" \"build.openshift.io.jenkins\" with name \"\" in project \"test\""}
```

That looks like a ref to the `build.openshift.io` api group and `jenkins` resource you mentioned before.

I added the `admin` role to the test user and logged out / logged back in, but same error.

Wasn't optimistic but I tried with no luck `oadm groups add-user ..` and either `build.openshift.io` or `build.openshift.io.jenkins` ... so as suspected api group does not correspond to `oadm groups`.

What is the next bit(s) to turn on here?  Should I go `user:full` ?

gabemontero: Get a state invalid / scope denied / 401 when attempt to add `user:full`

deads2k: The doc is really sticky on this type.  You need to specify an empty array for `scopes`, not `nil`.  Empty array says, "use all the powers of this user for the check".  `nil` says, "use the restriction of this token for the check".

PR #2, opened 2016-08-08
Get to work for me

gabemontero:
@smarterclayton PTAL
@bparees @deads2k FYI / PTAL if interested

The changes here are for:
1) fix pom version as part of publishing on jenkinsci.org
2) add some debug tracing
3) address TODO around deriving defaults, but not exposing those derived values subsequently in the UI (same basic approach used in the openshift/jenkins-plugin code)

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -30,12 +30,19 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
-import java.security.KeyStore.LoadStoreParameter;
 import java.security.KeyStoreException;
-import java.security.cert.CertificateFactory;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import javax.servlet.ServletException;
 
 import org.acegisecurity.Authentication;
 import org.acegisecurity.AuthenticationException;
@@ -58,6 +65,7 @@ import com.google.api.client.auth.oauth2.ClientParametersAuthentication;
 import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.auth.openidconnect.IdTokenResponse;
 import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpContent;
 import com.google.api.client.http.HttpRequest;
 import com.google.api.client.http.HttpRequestFactory;
 import com.google.api.client.http.HttpRequestInitializer;
@@ -68,11 +76,23 @@ import com.google.api.client.json.JsonObjectParser;
 import com.google.api.client.json.jackson2.JacksonFactory;
 import com.google.api.client.util.SecurityUtils;
 
+import hudson.EnvVars;
 import hudson.Extension;
 import hudson.Util;
+import hudson.model.Item;
+import hudson.model.Computer;
 import hudson.model.Descriptor;
+import hudson.model.Hudson;
+import hudson.model.Run;
 import hudson.model.User;
+import hudson.model.View;
+import hudson.scm.SCM;
+import hudson.security.Permission;
+import hudson.security.PermissionGroup;
+import hudson.security.GlobalMatrixAuthorizationStrategy;
+import hudson.security.ProjectMatrixAuthorizationStrategy;
 import hudson.security.SecurityRealm;
+import hudson.util.FormValidation;
 import hudson.util.HttpResponses;
 import hudson.util.Secret;
 import jenkins.model.Jenkins;
@@ -82,11 +102,32 @@ import jenkins.model.Jenkins;
  *
  */
 public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
+	static final Logger LOGGER = Logger.getLogger(OpenShiftOAuth2SecurityRealm.class.getName());
 
     /**
      * OAuth 2 scope. This is enough to call a variety of userinfo api's.
      */
-    private static final String SCOPE = "user:info";
+    private static final String SCOPE_INFO = "user:info";
+    private static final String SCOPE_CHECK_ACCESS = "user:check-access";
+    
+    static final String DEFAULT_SVC_ACCT_DIR = "/run/secrets/kubernetes.io/serviceaccount";
+    static final String DEFAULT_SVR_PREFIX = "https://openshift.default.svc";
+    static final String NAMESPACE = "namespace";
+    private static final String TOKEN = "token";
+    private static final String FINISH_METHOD = "doFinishLogin";
+    private static final String START_METHOD = "doCommenceLogin";
+    private static final String DISPLAY_NAME = "Login with OpenShift";
+    private static final String LOGIN_URL = "securityRealm/commenceLogin";
+    
+    private static final String[] ROLES = new String[]{"admin", "edit", "view"};
+    private static final String[] CLASSIC_DEFAULT_JENKINS_USERS = new String[]{"admin", "system_builder", "view", "Anonymous"};
+    
+    private static final String USER_URI = "/oapi/v1/users/~";
+    private static final String SAR_URI = "/oapi/v1/subjectaccessreviews";
+    private static final String OAUTH_PROVIDER_URI = "/.well-known/oauth-authorization-server";
+    
+    private static final String K8S_HOST_ENV_VAR = "KUBERNETES_SERVICE_HOST";
+    private static final String K8S_PORT_ENV_VAR = "KUBERNETES_SERVICE_PORT";
 
     /**
      * Global instance of the JSON factory.
@@ -94,7 +135,14 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
     private static final JsonFactory JSON_FACTORY = new JacksonFactory();
 
     static final HttpTransport HTTP_TRANSPORT = new NetHttpTransport();
-
+    
+    private static final Object USER_UPDATE_LOCK = new Object();
+    
+    private static List<String> classicDefaultJenkinsUsers = new ArrayList<String>();
+    static {
+    	classicDefaultJenkinsUsers = Arrays.asList(CLASSIC_DEFAULT_JENKINS_USERS);
+    }
+    
     /**
      * Control the redirection URL for this realm. Exposed for testing.
      */
@@ -107,7 +155,7 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
     /**
      * The transport loaded with a service account secret.
      */
-    private final HttpTransport transport;
+    private HttpTransport transport;
     
     /**
      * The service account directory, if set, instructs the plugin to follow the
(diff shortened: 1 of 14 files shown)


src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -132,34 +146,57 @@
...
+        this.serviceAccountName = Util.fixEmpty(serviceAccountName);
+        
+    	if (fixedServiceAccountDirectory == null && fixedClientId == null && fixedClientSecret == null) {
+    		overrides.put(SA_DIR, DEFAULT_SVC_ACCT_DIR);
+    	}
+    	this.serviceAccountDirectory = fixedServiceAccountDirectory;
+    	
+    	
+        if (fixedServerPrefix == null) {
+        	overrides.put(SVR_PREF, "https://openshift.default.svc");
```
smarterclayton: Why is this not also a constant?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -132,34 +146,57 @@
...
+    		if (fixedClientSecret == null) {
+	            BufferedReader r = new BufferedReader(new FileReader(new File(getServiceAccountDirectory(overrides), "token")));
+	            overrides.put(C_SEC, r.readLine());
+	            LOGGER.fine(String.format("ctor secret from file %s", overrides.get(C_SEC)));
 	            r.close();
     		}
-            if (Util.fixEmpty(clientId) == null) {
-	            BufferedReader r = new BufferedReader(new FileReader(new File(this.serviceAccountDirectory, "namespace")));
+            if (fixedClientId == null) {
+	            BufferedReader r = new BufferedReader(new FileReader(new File(getServiceAccountDirectory(overrides), "namespace")));
```
smarterclayton: Constant

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -132,34 +146,57 @@
...
+    	this.serviceAccountDirectory = fixedServiceAccountDirectory;
+    	
+    	
+        if (fixedServerPrefix == null) {
+        	overrides.put(SVR_PREF, "https://openshift.default.svc");
+        }
+    	if (overrides.containsKey(SA_DIR) || this.serviceAccountDirectory != null) {
+    		if (fixedClientSecret == null) {
+	            BufferedReader r = new BufferedReader(new FileReader(new File(getServiceAccountDirectory(overrides), "token")));
+	            overrides.put(C_SEC, r.readLine());
```
smarterclayton: No need to have overly short names for this - recommend CLIENT_SECRET_KEY_NAME

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -186,6 +225,18 @@ public OpenShiftOAuth2SecurityRealm(String serviceAccountDirectory, String servi
     public String getServiceAccountDirectory() {
         return serviceAccountDirectory;
     }
+    
+    /**
+     * For fields we derive defaults for when those fields are not set, those fields should be accessed via the getting
```
smarterclayton: I don't understand this godoc

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -132,34 +146,57 @@
      * The client secret from the OpenShift server.
      */
     private final Secret clientSecret;
+    
+    /**
+     * This map contains defaults gathered for this objects fields when those fields were not set in the UI.
+     * We don't store our inferred defaults in those fields, because if we do, they will show up in the gui
+     * When the getters are called.
+     */
+    private final Map<String,String> overrides = new HashMap<String,String>();
```
smarterclayton: It's questionable to me to create a string map like this.  Why not just make a private static type that has the same fields and use it consistently?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -232,19 +305,25 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+    	LOGGER.fine(String.format("newOAuthSession from %s redirectOnFinish %s", from, redirectOnFinish));
+    	final GenericUrl tokenServerURL = new GenericUrl(getServerPrefix(overrides) + "/oauth/token");
+        final String authorizationServerURL = getServerPrefix(overrides) +"/oauth/authorize";
+        
+        LOGGER.fine(String.format("tokenServerURL %s auth server url %s clientid %s clientSecret %s transport %s", tokenServerURL, authorizationServerURL, getClientId(overrides), getClientSecret(overrides) == null ? null : getClientSecret(overrides).getPlainText(), transport == null ? null : transport.toString()));
         
         final AuthorizationCodeFlow flow = new AuthorizationCodeFlow.Builder(
                 BearerToken.queryParameterAccessMethod(), transport, JSON_FACTORY, tokenServerURL,
-                new ClientParametersAuthentication(clientId, clientSecret.getPlainText()), clientId, authorizationServerURL)
+                new ClientParametersAuthentication(getClientId(overrides), getClientSecret(overrides).getPlainText()), getClientId(overrides), authorizationServerURL)
```
smarterclayton: Is this a common pattern in Jenkins?  It seems a violation of the object encapsulation pattern, and seems arbitrary.  Why not `getDefaultedClientId` instead and make it strongly typed?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -307,6 +388,7 @@ private String buildOAuthRedirectUrl() {
      * This is where the user comes back to at the end of the OpenID redirect ping-pong.
      */
     public HttpResponse doFinishLogin(StaplerRequest request) throws IOException {
+    	LOGGER.fine("Realm do finish " + request);
```
smarterclayton: Are logging messages supposed to prefix by the method that is calling them?  It's inconsistently applied in your debugging.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -315,5 +397,44 @@ public HttpResponse doFinishLogin(StaplerRequest request) throws IOException {
         public String getDisplayName() {
             return "Login with OpenShift";
         }
+
+        private FormValidation paramsWithViableDefaults(@QueryParameter String value) {
```
smarterclayton: `Viable` is unusual, suggest `paramsWithPodDefaults`

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -82,11 +89,18 @@
...
+	private static final Logger LOGGER = Logger.getLogger(OpenShiftOAuth2SecurityRealm.class.getName());
 
     /**
      * OAuth 2 scope. This is enough to call a variety of userinfo api's.
      */
     private static final String SCOPE = "user:info";
+    
+    private static final String DEFAULT_SVC_ACCT_DIR = "/run/secrets/kubernetes.io/serviceaccount";
+    private static final String SA_DIR = "serviceAccountDirectory";
+    private static final String C_ID = "clientId";
```
deads2k: `C` isn't something I auto-complete to client.  Can I get this one spelled? (the other abbreviations work in my head)

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -132,34 +146,57 @@
...
+     * When the getters are called.
+     */
+    private final Map<String,String> overrides = new HashMap<String,String>();
+    
 
     @DataBoundConstructor
     public OpenShiftOAuth2SecurityRealm(String serviceAccountDirectory, String serviceAccountName, String serverPrefix, String clientId, String clientSecret) throws IOException, GeneralSecurityException {
         HttpTransport transport = HTTP_TRANSPORT;
+        
+        LOGGER.fine(String.format("ctor args sa dir %s sa name %s svr prefix %s client id %s client secret %s", serviceAccountDirectory, serviceAccountName, serverPrefix, clientId, clientSecret));
```
deads2k: `isloggable` checks are best practice.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
deads2k: > Why is this not also a constant?

This isn't going to work for anyone.  Won't this be a source of annoyance and difficulty knowing where an how to fix it?  

I know the default was pre-existing, but I still question its value.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -132,34 +146,57 @@
...
-	            clientSecret = r.readLine();
+        String fixedServiceAccountDirectory = Util.fixEmpty(serviceAccountDirectory);
+        String fixedClientId = Util.fixEmpty(clientId);
+        String fixedClientSecret = Util.fixEmpty(clientSecret);
+        String fixedServerPrefix = Util.fixEmpty(serverPrefix);
+        
+        // decided not to default this since it is not even consistent within OpenShift;
+        // going through the example instantapp templates, the SA will be "default";
+        // going through the auto generated server, the SA will be "jenkins"
+        this.serviceAccountName = Util.fixEmpty(serviceAccountName);
```
deads2k: Discover this value by running an `oapi/v1/user/~` with the token.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
deads2k: > No need to have overly short names for this - recommend CLIENT_SECRET_KEY_NAME

I took a picture of this comment for posterity.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
deads2k: > Are logging messages supposed to prefix by the method that is calling them? It's inconsistently applied in your debugging.

Best practice is usually:
```java
if logger.isloggable(level) logger.logp(level, class, method, format, args)
```
That does a cheap check before building args for the logging method.  And the logging method doesn't have to perform an expensive stack inspection to determine the caller class/function.

If all you want is a, "did I call" point, you can use `logger.entering`.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: On Mon, Aug 8, 2016 at 9:16 PM, Clayton Coleman <notifications@github.com>
wrote:

> In src/main/java/org/openshift/jenkins/plugins/openshiftlogin/
> OpenShiftOAuth2SecurityRealm.java
> <https://github.com/openshift/jenkins-openshift-login-plugin/pull/2#discussion_r73984634>
> :
>
> > +        String fixedClientSecret = Util.fixEmpty(clientSecret);
> > +        String fixedServerPrefix = Util.fixEmpty(serverPrefix);
> > +
> > +        // decided not to default this since it is not even consistent within OpenShift;
> > +        // going through the example instantapp templates, the SA will be "default";
> > +        // going through the auto generated server, the SA will be "jenkins"
> > +        this.serviceAccountName = Util.fixEmpty(serviceAccountName);
> > +
> > +    	if (fixedServiceAccountDirectory == null && fixedClientId == null && fixedClientSecret == null) {
> > +    		overrides.put(SA_DIR, DEFAULT_SVC_ACCT_DIR);
> > +    	}
> > +    	this.serviceAccountDirectory = fixedServiceAccountDirectory;
> > +    	
> > +    	
> > +        if (fixedServerPrefix == null) {
> > +        	overrides.put(SVR_PREF, "https://openshift.default.svc");
>
> Why is this not also a constant?
>
It was originally a string literal, but I missed converting to a constant.
If we keep (re: subsequent comment from @deads2k), I'll convert.

gabemontero: I see references to `openshift.default.svc` and `kubernetes.default.svc` in the openshift code.  Is it a concern about the DNS being set up correctly?  In general I'm good with removing, just trying to understand the particulars.  Thanks.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: IMO, there is not a truly common pattern in Jenkins for this.  There is some "macro" related plugins wrt jenkinst global/job env vars and build steps (where there could be multiple instances of these build steps), but I have not seen much in the way of a common pattern where you discern defaults outside of those specific scopes.

The map based pattern for parameter storage stems from conventions on previous teams I was on ... old habits :-) ...

I'm good with moving to specific gets ... will move to that.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
deads2k: > I see references to openshift.default.svc and kubernetes.default.svc in the openshift code. Is it a concern about the DNS being set up correctly? In general I'm good with removing, just trying to understand the particulars. Thanks.

This name is used for two purposes:
 1. jenkins communicating with the master to exchange the authorization code for the access token.  The internal dns name works for that.
 2. redirect issued to the user to point them to the openshift master to authenticate to our oauth server.  It doesn't work for that.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: I'll add the if checks, but some background (I had specifically omitted them):
- there was actually a fair amount of debate back with our former employer's app server team and its performance group (@bparees sanity checked me on this, recalls this as well) wrt JIT optimizations, 2 function calls vs. one when trace was enabled, cluttering the code for non-critical performance paths, along with the points you rose)
- There general opinion leaned toward omitting the if checks, but I think it is fair to say that was not conclusive

Aside from adding the if check, I'm also reducing the amount of trace points (in hindsight a good portion of it stemmed from being new to the code and oauth flow and trying to get things working for the first time) and will clean up the prefixing.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: I've tried a few permuations:
- list users
- get user "~"
- get user "test" (where I logged in as test previously)

And I keep getting refused because the service account does not have permission to get users.

Typically, we grant the serviceaccount jenkins is running under the "edit" role.

I've even gone as far as adding the "sudoer" role to the serviceaccount, but something is still not right.

What aspect(s) do I need to change?

gabemontero: `oc rsh` into the jenkins pod, and and `oc get user ..` failed as well.  

Trolling through clusterPolicy, clusterPolicyBindings, and running `oc policy add-role-to-user ... -z default`, seeing if one of those opens the gate.

Course corrections welcome though - thanks.

deads2k: `oc get user ~` should always work in pod.  Can you post `--loglevel=8`?

gabemontero: OK, `oc get user ~` did not work, but `oc get user/~` does.  That said, the error I was getting when trying in Java was not the same as the command line.  I still get permission denied in Java.  But using the `--loglevel=8` output, I should be able to work backwards and get something to work.  Might have to ditch the openshift-restclient-java in the process and go with lower level okhttp packages, but no biggie.

Thanks.

gabemontero:
responses to comments pushed

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
smarterclayton: This is how jenkins communicates with the master - so it will always work
inside the cluster.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
smarterclayton: Constants should generally be descriptive.  Method names should be a bit
more terse without introducing shortcuts.  Variables can be very terse.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
smarterclayton: I'm asking more within the context of Jenkins plugins.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
smarterclayton: oc get user ~ does not work because the shell expands ~ to $HOME.

gabemontero: ahhhhhh

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: My comments about defaulting fields include the context of Jenkins plugins.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: Yeah, but the browser logging into jenkins may not be inside the cluster.  That is @deads2k point #2, no?

To confirm, I just tried it, and got from firefox:  `Firefox can't find the server at openshift.default.svc.`

gabemontero:
@smarterclayton @deads2k is there still disagreement on no longer using `https://openshift.default.svc` as a default?

if no disagreement, bump

gabemontero:
OK, aside from some tweaking base on iterations of testing / examination, I've put together changes along the lines of what is proposed in @deads2k 's https://github.com/openshift/jenkins-openshift-login-plugin/pull/1

Also, I think we still need confirmation from @smarterclayton that he is in agreement with @deads2k and I about removing the use of `https://openshift.default.svc` as a default for the server prefix.

PTAL - thanks

smarterclayton:
We should split the prefix then.  Is this is in a working state that's fine and we can come back later.

gabemontero:
On Fri, Aug 19, 2016 at 6:46 PM, Clayton Coleman <notifications@github.com>
wrote:

> We should split the prefix then. Is this is in a working state that's fine
> and we can come back later.
>
It is in a working state.  I'm not following what you mean by split the
prefix.

deads2k:
> It is in a working state.  I'm not following what you mean by split the
> prefix.

He's not convinced to remove the default.  Leave it as-is for now to merge the rest of your pull and argue over the server prefix in a different pull/issue.

gabemontero:
Roger that

smarterclayton:
Also by split the prefix I meant split into the internal prefix (for API
calls) and the external prefix (for redirects)

gabemontero:
On Mon, Aug 22, 2016 at 1:08 PM, Clayton Coleman <notifications@github.com>
wrote:

> Also by split the prefix I meant split into the internal prefix (for API
> calls) and the external prefix (for redirects)
>

Ah - yep, 10-4


> On Mon, Aug 22, 2016 at 8:23 AM, Gabe Montero <notifications@github.com>
> wrote:
>
> > Roger that
> >

gabemontero:
I still have to add some more tests, but have pushed an update in the interim because of changes induced by more manual testing (in addition to responses to past iteration of comments).

This go around:
- no longer hard code json string for self-sar

- cleaned up error handling around stale http session, missing oauth session attribute; now redirect to root jenkins url to force re-login if in this state; aligns with behavior with default security realm, where it forces you to log back in if your session has expired

- split out api call url and redirect url into separate fields, separate defaults;  api calls default to "https://openshift.default.svc"; redirect url defaults, at least for now, to the api svc cluster IP, which is available via the env vars set in the pod and picked up by Jenkins ... so for demo'ing on a single, local host, you don't have to specify any config in Jenkins to get it to work

- backed off the technique that @deads2k outlined in https://github.com/openshift/jenkins-openshift-login-plugin/pull/1/files#diff-d68a65b0daea9ede707036e33ba65f4aR267 ; after a bunch of testing, experimentation, tracing in core Jenkins, I could not get Jenkins's ACLs to validate users against the Jenkins default Permission set.  It seems like it should work, but it doesn't.  My interpretation, based on the tenor of the security documentation, the AuthorizationStrategy plug point is the one that gets more focus and, not getting out of the way of this pun, more strategic.

I have duplicated what the Jenkins Authorization/Access Control UI under Configure Global Security does when it adds users or modifies their permissions.  With this change, we now avoid the need to add users manually with the minimal amount of permissions needed to fully log in.  If they authenticate against OpenShift OAuth, they'll get added based on the results of the self-sar with permission sets currently (rouglly) analogous to our view/edit/admin roles.  Certainly the mapping can be massaged, especially as the build.io.openshift api group / jenkins resource stuff that @deads2k mentioned in his pull is fully realized.  Btw, if I'm misreading current state of affairs, and that stuff is fully realized already, then I may need another round of assistance with the self-sar stuff.

I also have some //TODO's around a few question areas, like minimum roles allowed and what to do with the current set of users defined in our Jenkins image (@bparees and I have discussed this some on the side)

PTAL @smarterclayton @deads2k @bparees

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -232,43 +481,110 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+                        	newAuthMgr.add(Item.READ, info.getName());
+                			if (allowedRoles.contains("edit") || allowedRoles.contains("admin")) {
+                				newAuthMgr.add(Item.BUILD, info.getName());
+                				newAuthMgr.add(Item.CONFIGURE, info.getName());
+                				newAuthMgr.add(Item.CREATE, info.getName());
+                				newAuthMgr.add(Item.DELETE, info.getName());
+                				newAuthMgr.add(Item.WORKSPACE, info.getName());
+                				newAuthMgr.add(SCM.TAG, info.getName());
+                			}
+                			if (allowedRoles.contains("admin")) {
```
deads2k: Why isn't there a "view" stanza here?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -232,43 +481,110 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+                				newAuthMgr.add(Hudson.READ, info.getName());
+                				newAuthMgr.add(Run.DELETE, info.getName());
+                				newAuthMgr.add(Run.UPDATE, info.getName());
+                				newAuthMgr.add(View.CONFIGURE, info.getName());
+                				newAuthMgr.add(View.CREATE, info.getName());
+                				newAuthMgr.add(View.DELETE, info.getName());
+                			}
+                    	}
+            			
+            			Jenkins.getInstance().setAuthorizationStrategy(newAuthMgr);
```
deads2k: If a user loses access later on, does this clean up after them to revoke their access?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: I'm assuming (per the comment above with the TODO / question) that everyone
will have at least "view" access.
Please correct me if that is not the case.  If it is not the case, the next
step will be to confirm what is needed to at least
log in successfully (minimally need Hudson.READ for sure; will have to try
with and without Item.READ) vs. what
else might we want to add that would be analogous for the "view" role in
OpenShift.

The "view" user predefined in our Jenkins image today has Hudson.READ and
Item.READ.

deads2k: > I'm assuming (per the comment above with the TODO / question) that everyone
> will have at least "view" access.
> Please correct me if that is not the case.  If it is not the case, the next
> step will be to confirm what is needed to at least
> log in successfully (minimally need Hudson.READ for sure; will have to try
> with and without Item.READ) vs. what
> else might we want to add that would be analogous for the "view" role in
> OpenShift.
> 
> The "view" user predefined in our Jenkins image today has Hudson.READ and
> Item.READ.

No.  I'll work up a pull to create the role to add jenkins permissions (based on these values), but most users will have no rights to a random jenkins instance.

deads2k:
Opened https://github.com/openshift/origin/pull/10649 to enable these permissions.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: On Thu, Aug 25, 2016 at 2:48 PM, David Eads <notifications@github.com>
wrote:

> In src/main/java/org/openshift/jenkins/plugins/openshiftlogin/
> OpenShiftOAuth2SecurityRealm.java
> <https://github.com/openshift/jenkins-openshift-login-plugin/pull/2#discussion_r76302766>
> :
>
> > +                				newAuthMgr.add(SCM.TAG, info.getName());
> > +                			}
> > +                			if (allowedRoles.contains("admin")) {
> > +                				newAuthMgr.add(Computer.CONFIGURE, info.getName());
> > +                				newAuthMgr.add(Computer.DELETE, info.getName());
> > +                				newAuthMgr.add(Hudson.ADMINISTER, info.getName());
> > +                				newAuthMgr.add(Hudson.READ, info.getName());
> > +                				newAuthMgr.add(Run.DELETE, info.getName());
> > +                				newAuthMgr.add(Run.UPDATE, info.getName());
> > +                				newAuthMgr.add(View.CONFIGURE, info.getName());
> > +                				newAuthMgr.add(View.CREATE, info.getName());
> > +                				newAuthMgr.add(View.DELETE, info.getName());
> > +                			}
> > +                    	}
> > +            			
> > +            			Jenkins.getInstance().setAuthorizationStrategy(newAuthMgr);
>
> If a user loses access later on, does this clean up after them to revoke
> their access?
>
The next time the user logs in and creates a new session, yes, the new
permissions should be reflected.  They'll maintain the privileges they
possessed
when the logged into Jenkins for the life of that session currently.  In
theory we could add  background polling to augment things while the user's
session is active
but my initial thought is that is excessive (but let me know if you
disagree).

deads2k: > when the logged into Jenkins for the life of that session currently.  In
> theory we could add  background polling to augment things while the user's
> session is active
> but my initial thought is that is excessive (but let me know if you
> disagree).

How long do sessions last?  If they're short term (say <1 day), then we probably don't need to double check.  If they're long term (say >1 day), we'll probably want something eventually.

gabemontero: They  go with the servlet spec default - 30 minutes.

gabemontero: Meaning if inactive for 30 minutes, timeout.  If they are continually doing
stuff, the session stays active.

deads2k: > Meaning if inactive for 30 minutes, timeout.  If they are continually doing
> stuff, the session stays active.

Is there a jenkins option to boot all sessions after X duration?

gabemontero: We had an option like that in the app server :-) but it is not standard
servlet spec.  I'll poke around a bit
to make sure but I doubt Jenkins (or the servlet engine it uses under the
covers) has it.

gabemontero: More of the servlet api came back to me.  We should be able to register a HttpSessionAttributeListener that includes call outs when the session is accessed.  We should be able to leverage that callout to poll openshift for updated permissions.  And then if they don't access the session, it should time out and the refresh will occur when the log back in.

I'll whip something up along those lines tomorrow.

gabemontero: I deleted the previous brainstorm on http session related events ... refreshed my mem with the java doc.  None of the 3 listeners are called on each session access.  The callouts present won't serve our periodic polling needs.

A servlet filter would work, but you specify those in the web.xml.  I see some indirect refs to tweaking their web.xml in stackoverflow, but nothing that gives me confidence it is something we should be doing.

If refresh in between sessions is not sufficient, we are back to polling in a background thread, or polling say as part of the pipeline plugin's operation.  There are standard Jenkins api that would allow it to get the openshift oauth login plugin.

deads2k: @liggitt See this chain about re-detection of changed permissions.  I'm concerned about ability for a long term session to never have its permissions re-checked.  Are you comfortable with this as-is and simply leaving it up to an administrator to find a way to boot live sessions (I don't know how to do this jenkins)?  I'd prefer to have the permissions redetected at some sane interval, but it doesn't sound like an easy thing to do.

gabemontero: OK, in finding [these monitoring scripts](https://wiki.jenkins-ci.org/display/JENKINS/Monitoring+Scripts) and looking at its [source code](https://github.com/jenkinsci/monitoring-plugin), I discovered that Jenkins has a [programmatic hook](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/util/PluginServletFilter.java#L98) for adding servlet filters.

So I should be able to re-validate a users permissions periodically (even on every request, thought that seems excessive) while still active.

I'll update the PR when I have the change ready.

gabemontero: OK, I've pushed the change with the servlet filter so that we poll openshift for updated permissions as the given user traverses through jenkins.

I've initially declined to expose the polling interval as a config param.  My hope is there is an agreeable default and the user has 1 less tuning know to digest.  But of course chime in as desired.

smarterclayton: I'm a little concerned about refreshing tied to the session.  I don't like
polling, and I'm not sure that we need rapid reaction to permissions in
openshift.  What other options do we have?

deads2k:
> I'm a little concerned about refreshing tied to the session.  I don't like
polling, and I'm not sure that we need rapid reaction to permissions in
openshift.  What other options do we have?

@smarterclayton To be clear, this is not about rapid reaction, this about *any* reaction.  Without any changes, you log in once, have access, keep the window open (clicking periodically), and you *never* lose those powers.  I also asked in the chain about whether an admin had any option to boot the sessions, but that didn't seem like something that could be done either.

gabemontero:
On Monday, August 29, 2016, David Eads <notifications@github.com> wrote:

> I'm a little concerned about refreshing tied to the session. I don't like
> polling, and I'm not sure that we need rapid reaction to permissions in
> openshift. What other options do we have?
>
> @smarterclayton <https://github.com/smarterclayton> To be clear, this is
> not about rapid reaction, this about *any* reaction. Without any changes,
> you log in once, have access, keep the window open (clicking periodically),
> and you *never* lose those powers. I also asked in the chain about
> whether an admin had any option to boot the sessions, but that didn't seem
> like something that could be done either.
>
The admin doesn't have an invalidate all sessions button in the gui but we
could invalidate sessions in this filter on various scheduling notions
(certain times of days or amount of time the session has been active).

There are cons to that as well, and we start to get into user experience
concerns vs infrastructure concerns.

I at least don't know where I land yet on one vs the other.

bparees:
@gabemontero @smarterclayton @deads2k I definitely have concerns about the user experience if we're logging users out periodically, and I can't see us willingly logging users out more often than once a day (without really annoying them), at which point it's of dubious value from a security perspective anyway.

It's been a while since i've done oauth flows, is there any way to re-check the permissions under the covers w/o user interaction?

bparees:
(or is that just another way of saying polling?)

deads2k:
> day (without really annoying them), at which point it's of dubious value from a security perspective anyway.

In terms of a permissions recheck, once a day is a tremendous improvement over never.

gabemontero:
On Mon, Aug 29, 2016 at 10:42 AM, David Eads <notifications@github.com>
wrote:

> day (without really annoying them), at which point it's of dubious value
> from a security perspective anyway.
>
> In terms of a permissions recheck, once a day is a tremendous improvement
> over never.
>
I've left the polling in, but changed the interval to poll to 24 hours.

smarterclayton:
Wait, but doesn't the oauth token have a lifetime of 24 hours?  Can't we
just check then?

gabemontero:
If the oauth token has a event listener API around when the lifetime
expires, then sure, I
can create / register one.

I'll dig a bit, see if one exists, but any detailed pointers you've got off
the top of your head
are certainly welcome.

gabemontero:
Or, wait, are you saying that we'll have to re-login anyway after 24 hours,
irregardless of how often we interact with Jenkins from the UI, maintain
the life of the http session, etc. ... is that what you mean?

smarterclayton:
Wait, can we not constrain Jenkins login sessions at all to a fixed time?
Because if we constrained it to the lifecycle of the token, that is a good
start.  I'm not totally opposed to a background poller (so need to take
it), I just want to be deliberate about it.

smarterclayton:
Yes I think so.  Maybe not, but that's somewhat expected.

gabemontero:
On Mon, Aug 29, 2016 at 4:33 PM, Clayton Coleman <notifications@github.com>
wrote:

> Wait, can we not constrain Jenkins login sessions at all to a fixed time?
> Because if we constrained it to the lifecycle of the token, that is a good
> start. I'm not totally opposed to a background poller (so need to take
> it), I just want to be deliberate about it.
>

Jenkins solely works off of the http session and the servlet spec here, so
it is an inactivity based timeout.
There are no fixed time duration option to leverage.

I'll sit tight and work on other aspects of the solution while you
deliberate.

gabemontero:
On Mon, Aug 29, 2016 at 4:33 PM, Clayton Coleman <notifications@github.com>
wrote:

> Yes I think so. Maybe not, but that's somewhat expected.
>

I see.  OK, we can circle back if any confirmations arise, or as we wrap up
the polling
deliberations.

deads2k:
> Wait, can we not constrain Jenkins login sessions at all to a fixed time?

This would work for me.  I just want a bounded time for a permission change to be reflected.  Forcing a logout would work fine.  Otherwise, you end up in cases where after changing a `ClusterRole`, you have to cycle every jenkins instance in the entire cluster.

gabemontero:
FYI @deads2k @smarterclayton @bparees - I updated my local openshift/origin with the changes from https://github.com/openshift/origin/pull/10649 and my self-sars against "build.openshift.io" and the "jenkins" resource are now coming back stating my users is approved for view, edit, admin as expected.

I've updated the pull with some cleanup around this.

The pull now also has some additional changes:

1) a slight increase in the jenkins version to match what is the minimal version specified in openshift-sync and openshift-pipeline.  Aside from consistency, this moves us closer down the path of upgrading the parent-pom to 2.x, which is something the jenkins community is definitely "pushing" at the moment.  Ultimately, I'd like to stay ahead of the inevitable point where they mandate such a bump to be able to host the plugins at the jenkins update center.  However, as various underlying dependencies change when you do this, moving the parent pom to 2.x is not a flip the switch change.  Additional tweaks to the unit tests and pom.xml appear required based on initial experiments.

The change to date already necessitated some tweaks to the unit tests, as htmlunit related dependencies underwent API changes, and I had to adapt in order to compile / run.

2) I've introduced changes so that the oauth plugin can be the default security realm on startup for our jenkins image.  I'm doing this programmatically instead of changing the config templates in the jenkins image.  The reasoning / logic is as follows:
  a) first see if we are already running with the openshift oauth login; if so, leave things as is
  b) if not, determine via inspection of the env vars and presence of files mounted in the pod that we are in fact in a pod
  c) if in a pod, change the security realm to the openshift oauth plugin, and use the defaults from the pod env and what is defined on the openshift cluster side wrt authenticate/authorization.
  d) if not in a pod, leave the default security realm in place; the user can login with that (admin/password, etc.), and then change to oauth if they want

3) I've added some config for a couple of elements of this, but rather than expose them through the Jenkins UI, I've introduced them solely as env vars which can be set when jenkins is deployed from the openshift side.  Those props are:
   a) "OPENSHIFT_DISABLE_SECURITY_REALM_OVERRIDE" - setting this to anything no-op's 2)
   b) "OPENSHIFT_POLL_INTERVAL_OVERRIDE" - the interval for polling for permissions (default 1 day) that we've recently discussed

bparees:
> a) "OPENSHIFT_DISABLE_SECURITY_REALM_OVERRIDE" - setting this to anything no-op's 2)
> b) "OPENSHIFT_POLL_INTERVAL_OVERRIDE" - the interval for polling for permissions (default 1 day) that we've recently discussed

probably don't need "OVERRIDE" in either of these names.  for most of our images we have default behavior that gets overridden if you set some env variable.

and for (a) we probably want a more meaningful name for users like OPENSHIFT_DISABLE_OAUTH or DISABLE_OPENSHIFT_OAUTH or something that makes it clearer to users the implications.

for (b) it should probably indicate what it's polling, eg OPENSHIFT_PERMISSIONS_POLL_INTERVAL or something.

gabemontero:
On Wed, Aug 31, 2016 at 1:51 PM, Ben Parees <notifications@github.com>
wrote:

> a) "OPENSHIFT_DISABLE_SECURITY_REALM_OVERRIDE" - setting this to anything
> no-op's 2)
> b) "OPENSHIFT_POLL_INTERVAL_OVERRIDE" - the interval for polling for
> permissions (default 1 day) that we've recently discussed
>
> probably don't need "OVERRIDE" in either of these names. for most of our
> images we have default behavior that gets overridden if you set some env
> variable.
>
> and for (a) we probably want a more meaningful name for users like
> OPENSHIFT_DISABLE_OAUTH or DISABLE_OPENSHIFT_OAUTH or something that makes
> it clearer to users the implications.
>
Updated to OPENSHIFT_DISABLE_OAUTH

> for (b) it should probably indicate what it's polling, eg
> OPENSHIFT_PERMISSIONS_POLL_INTERVAL or something.
>

Updated to OPENSHIFT_PERMISSIONS_POLL_INTERVAL

gabemontero:
Just pushed an update to add some unit tests around defaulting params per prior request.

gabemontero:
Updated based on today's face-to-face discussion the following items:
1) the default polling interval is 5 minutes
2) Changed the env var to OPENSHIFT_ENABLE_OAUTH, and only override the sec realm on startup if that is set (where we will update the jenkins example ephemeral and persistent templates to set that env var)

jwforres:
@gabemontero screenshots of what this flow looks like?

gabemontero:
glad you asked @jwforres - much assistance wrt look and feel will be needed here :-)

Initial prompt:
prompt

If you click OK, you get the OpenShift login screen:
login

If you click Cancel, you get this simple message:
cancel

I can send you the *.png files as well if desired.

gabemontero:
I've pushed updates from further testing as well as consuming the new capabilities provided by https://github.com/openshift/origin/pull/10845 and https://github.com/openshift/origin/pull/10878.

With these changes and specifying 2 generic (meaning no specific ip's / hostnames) annotations on the jenkins service account specified in the jenkins example templates, a user can log into jenkins via openshift oauth with no manual steps whatsoever using `oc cluster up`.

In theory, once the two origin PRs merge, this PR could merge and we could gen up a new version of this plugin and pull it into the jenkins image.

That said, I still need to circle back with @jwforres on getting a more official login page from her team, but that could happen in a separate PR if we wanted to get QE started sooner rather than later.

PTAL and comment at your convenience.

smarterclayton:
Would like to get everything in and enabled by default and used ASAP, even
if not pretty.

PR #7, opened 2017-01-06
access jenkins via bearer token self-sar against openshift oauth; ens…

gabemontero:
…ure filter is active after restart

@bparees FYI / PTAL

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -125,7 +125,6 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
     private static final String LOGIN_URL = "securityRealm/commenceLogin";
     
     private static final String[] ROLES = new String[]{"admin", "edit", "view"};
-    private static final String[] CLASSIC_DEFAULT_JENKINS_USERS = new String[]{"admin", "system_builder", "view", "Anonymous"};
     
     private static final String USER_URI = "/oapi/v1/users/~";
     private static final String SAR_URI = "/oapi/v1/subjectaccessreviews";
@@ -146,11 +145,6 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
     static final HttpTransport HTTP_TRANSPORT = new NetHttpTransport();
     
     private static final Object USER_UPDATE_LOCK = new Object();
-    
-    private static List<String> classicDefaultJenkinsUsers = new ArrayList<String>();
-    static {
-    	classicDefaultJenkinsUsers = Arrays.asList(CLASSIC_DEFAULT_JENKINS_USERS);
-    }
     
     /**
      * Control the redirection URL for this realm. Exposed for testing.
@@ -246,6 +240,8 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
      */
     private OpenShiftProviderInfo provider;
     
+    private OpenShiftPermissionFilter filter;
+    
     @DataBoundConstructor
     public OpenShiftOAuth2SecurityRealm(String serviceAccountDirectory, String serviceAccountName, String serverPrefix, String clientId, String clientSecret, String redirectURL) throws IOException, GeneralSecurityException {
         HttpTransport transport = HTTP_TRANSPORT;
@@ -277,21 +273,36 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
         else
         	populateDefaults();
     	
-        try {
-			hudson.util.PluginServletFilter.addFilter(new OpenShiftPermissionFilter());
-		} catch (ServletException e) {
-			LOGGER.log(Level.SEVERE, "ctor", e);
-		}
-        
         if (LOGGER.isLoggable(Level.FINE))
         	LOGGER.fine(String.format("ctor: derived default client id %s client secret %s sa dir %s transport %s", defaultedClientId, defaultedClientSecret, defaultedServiceAccountDirectory, transport));
     }
     
+    /*
+     * Note, a fair amount of investigation was done into leveraging the hudson.security.SecurityRealm.createFilter(FilterConfig) extension
+     * point.  However, quite a bit of tight coupling with assumed behavior in the jenkins core wrt the other security features and servlet filtering
+     * arose.  Hence, we are sticking with the hudson.util.PluginServletFilter.addFilter(Filter) path.
+     */
+    synchronized void createFilter() {
+        // restarts on things like plugin upgrade bypassed the call to the constructor, so filter initialization
+        // has to be driven in-line; note, after initial bring up, the filter variable will be set after subsequent
+        // jenkins restarts, but the addFilter call needs to be made on each restart, so we check flag to see if the filter
+        // has been ran through at least once
+        if (filter == null || !filter.initCalled) {
+            try {
+                filter = new OpenShiftPermissionFilter();
+                hudson.util.PluginServletFilter.addFilter(filter);
+            } catch (ServletException e) {
+                LOGGER.log(Level.SEVERE, "createFilter", e);
+            }
+        }
+    }
+    
     boolean populateDefaults() throws IOException, GeneralSecurityException {
-    	boolean runningInOpenShiftPodWithRequiredOAuthFeatures = EnvVars.masterEnvVars.get(K8S_HOST_ENV_VAR) != null && EnvVars.masterEnvVars.get(K8S_PORT_ENV_VAR) != null;
+        createFilter();
+        boolean runningInOpenShiftPodWithRequiredOAuthFeatures = EnvVars.masterEnvVars.get(K8S_HOST_ENV_VAR) != null && EnvVars.masterEnvVars.get(K8S_PORT_ENV_VAR) != null;
         // we want to be verbose wrt error logging if we are at least running within a pod ... but if we know we are outside a pod, only
-    	// log if trace enabled 
-    	boolean withinAPod = runningInOpenShiftPodWithRequiredOAuthFeatures || (new File(getDefaultedServiceAccountDirectory())).exists();
+        // log if trace enabled 
+        boolean withinAPod = runningInOpenShiftPodWithRequiredOAuthFeatures || (new File(getDefaultedServiceAccountDirectory())).exists();
         
         FileInputStream fis = null;
         BufferedReader br = null;
@@ -437,10 +448,6 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
     public String getDefaultedNamespace() {
     	return namespace;
     }
-    /*
-     *     public OpenShiftOAuth2SecurityRealm(String serviceAccountDirectory, String serviceAccountName, String serverPrefix, String clientId, String clientSecret, String redirectURL) throws IOException, GeneralSecurityException {
-
-     */
     
     
 
@@ -591,7 +598,7 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
                             flow.newTokenRequest(authorizationCode).setRedirectUri(buildOAuthRedirectUrl(redirectOnFinish)));
                     final Credential credential = new Credential(BearerToken.authorizationHeaderAccessMethod()).setFromTokenResponse(response);
                 	this.setCredential(credential);
-                    secRealm.updateAuthorizationStrategy(this);
+                    secRealm.updateAuthorizationStrategy(credential);
                             			
                     return new HttpRedirect(redirectOnFinish);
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -277,17 +273,32 @@ public OpenShiftOAuth2SecurityRealm(String serviceAccountDirectory, String servi
...
+                filter = new OpenShiftPermissionFilter();
+                hudson.util.PluginServletFilter.addFilter(filter);
+            } catch (ServletException e) {
+                LOGGER.log(Level.SEVERE, "createFilter", e);
+            }
+        }
+    }
+    
     boolean populateDefaults() throws IOException, GeneralSecurityException {
+        createFilter();
```
bparees: indenting is misaligned

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -605,16 +612,35 @@ public HttpResponse onSuccess(String authorizationCode) {
...
+                
+        // we append the role suffix to the name stored into Jenkins, since a given user is able to log in at varying scope/permission
+        // levels in openshift; however, for now, we make sure the display name for Jenkins does not include this suffix
+        String suffix = "";
+        if (allowedRoles.contains("admin"))
+            suffix = "-admin";
+        else if (allowedRoles.contains("edit"))
+            suffix = "-edit";
+        else if (allowedRoles.contains("view"))
+            suffix = "view";
```
bparees: what about browser based logins?  I thought we were going to handle them with a unique suffix so they don't interfere with scoped token access.  (consider the case where i have both a scoped token that only has view access, and i login to the jenkins console w/ my regular user who also only has view access.  Then my permissions get updated to give my normal user admin access.  The polling logic is going to update the matrix to change my "user.view" account to have admin access).

"user.browser" or "user.dynamic" seem like good ways to distinguish those accounts.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftPermissionFilter.java
```diff
@@ -78,25 +122,58 @@ public void doFilter(ServletRequest request, ServletResponse response,
...
+	                        if (words.length > 1) {
+	                            String token = words[1];
+	                            
+	                            BearerCacheEntry entry = bearerCache.get(token);
+	                            if (updated || entry == null || System.currentTimeMillis() - entry.lastCheck.longValue() > (interval * 1000)) {
+	                                final Credential credential = new Credential(BearerToken.authorizationHeaderAccessMethod()).setAccessToken(token);
+	                                OpenShiftOAuth2SecurityRealm secRealm = (OpenShiftOAuth2SecurityRealm) Jenkins.getInstance().getSecurityRealm();
+	                                UsernamePasswordAuthenticationToken jenkinsToken = secRealm.updateAuthorizationStrategy(credential);
+	                                
+	                                //TODO can we assume that once a token is invalid, it is always invalid?  If so, we could put storage of the token
```
bparees: it won't "always" be invalid necessarily, but i think it's fair to cache it as being invalid for the same 30s(or whatever) that we are caching any other token auth check result.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: this path applies to both browser based and curl based logins.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: hmmmm ... its lined in mine and I thought my eclipse was converting everything to spaces.  I'll see what is up

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: Perhaps I should add comments to better clarify, but independent of browser or curl, a given user as defined in openshift can have up to 3 entries (each a unique permission level) into jenkin's auth matrix, where none of the keys for those entries is the user name defined in openshift.  The base user name used for the key is not necessarily what they might have supplied on the login, but whatever oauth self check on the token reports as the user name (where that token is supplied in the curl case, and is provided by the oauth server in the browser case).  

Then, that base user name is never stored in the Jenkins auth matrix.  Only <base user name>-admin, <base user name>-edit, <base user name>-view.

Then, based on the self-sar results, you get one of those 3 permission levels.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftPermissionFilter.java
gabemontero: The interval is configurable.  As an FYI, the default is 5 minutes (that is what I worked out with David re: how often we redo the self sar).

But taking your comment as I currently interpret it, I'll update to cache the invalid entries as well.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: Yep, some lines still had some tabs.  Perhaps they were set that way before I switched my eclipse.  Update part of next push.

gabemontero:
OK the indentation and caching invalid entries changes are pushed.  I deferred on potential clarifying comments for the suffix, etc. related thread until our discussion results in our understandings converging one way or the other.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
bparees: yes but what happens when the initial self-sar gives me view permission and you log me in as "myuser.view" and then on a polling update, you find that i now have admin access?  You can't change my username for the current sesssion(I assume), so you have to update the permissions associated w/ the user i'm currently logged in as (which is myuser.view), at which point you're in the race scenario we had discussed.

gabemontero: "myuser.view" is never your actual username.  It is only the key into the jenkins auth matrix, which is constructed based on your actual username ("myuser") and the self-sar results of your token.  The key *can* change for the current session as the permissions for username/token changes.

This level of indirection hence avoids the race scenario.

bparees: ah, ok, let me take another look through this w/ that understanding.  thanks for the explanation!

gabemontero: Cool - thanks!

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -605,16 +612,35 @@ public HttpResponse onSuccess(String authorizationCode) {
...
+        String matrixKey = info.getName() + suffix;
+        
         // logs this user in.
         UsernamePasswordAuthenticationToken token =
-                new UsernamePasswordAuthenticationToken(info.getName(), "", authorities);
+                new UsernamePasswordAuthenticationToken(matrixKey, "", authorities);
         SecurityContextHolder.getContext().setAuthentication(token);
         
+        User u = User.get(token.getName());
```
bparees: token.getName() is the name of the token, or the name of the account associated w/ the the token?  if it's the latter it's going to be a problem once we support scoped tokens and so you have multiple tokens for the same username.

(and if it's the former, it seems like it would be a problem if the same token name is used within different projects)

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -605,16 +612,35 @@ public HttpResponse onSuccess(String authorizationCode) {
...
+        info.updateProfile(u);
```
bparees: what's the relationship between "userinfo" and the profile object?  maybe that addresses my naming concern above.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftPermissionFilter.java
```diff
@@ -66,6 +105,15 @@ public void doFilter(ServletRequest request, ServletResponse response,
...
 			final HttpServletRequest httpRequest = (HttpServletRequest) request;
+            long interval = SELF_SAR_POLL_INTERVAL;
+            String var = EnvVars.masterEnvVars.get(OPENSHIFT_PERMISSIONS_POLL_INTERVAL);
+            if (var != null) {
+                try {
+                    interval = Long.parseLong(var);
+                } catch (Throwable t) {
+                    
+                }
+            }
```
bparees: indenting

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftPermissionFilter.java
```diff
@@ -78,25 +126,61 @@ public void doFilter(ServletRequest request, ServletResponse response,
...
 				}
+			} else {
+			    // support for non-browser, like curl, access to jenkins with openshift oauth security;
+			    // by choice, not storing auth in http session (remember, no browser) or anything like that;
+			    // want the token provided on each access 
+			    try {
+			        String enabled = EnvVars.masterEnvVars.get(OPENSHIFT_ACCESS_VIA_BEARER_TOKEN);
+			        if (enabled == null || !enabled.equalsIgnoreCase("false")) {
+	                    String authHdr = httpRequest.getHeader("Authorization");
+	                    if (authHdr != null && authHdr.length() > 0 && authHdr.startsWith("Bearer")) {
```
bparees: indenting

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftPermissionFilter.java
```diff
@@ -78,25 +126,61 @@ public void doFilter(ServletRequest request, ServletResponse response,
...
+	                        String[] words = authHdr.split(" ");
+	                        if (words.length > 1) {
+	                            String token = words[1];
+	                            
+	                            BearerCacheEntry entry = bearerCache.get(token);
+	                            if (entry == null) {
+                                    entry = new BearerCacheEntry();
+                                    bearerCache.put(token, entry);
```
bparees: seems slightly weird to put the entry in the cache before you've initialized it.  can this move down into the if block below?  (I realize due to code reordering that still may not be sufficient to avoid weird races)

gabemontero: the `UsernamePasswordAuthenticationToken jenkinsToken = secRealm.updateAuthorizationStrategy(credential);`, currently at line 161, will throw an exception if an invalid credential is passed in.  That eliminates putting the entry in after the if-then / else if / else blocks, and forces us to duplicate the put in the if-then block, else if block, and else block.

A cache entry without the token field set but still with the lastCheck field set has value for our logic.

Pending some nuance I'm missing that you highlight to me, given the duplication point above, I think we should keep as is.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftPermissionFilter.java
gabemontero: Yeah the eclipse/format the entire file does a bunch and refactoring I don't like, so i'll have to handle these manually.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftPermissionFilter.java
bparees: by "invalid credential" that includes credentials that reference tokens which do not exist in the cache?  that is, the lookup must succeed but it's ok that the lookup returns an entry w/o a valid token?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: It is the name of the account, if I gather what you mean by that correctly.  When I bring things up via `oc cluster up`, and log in as `developer`, `token.getName()` returns `developer`.  It is *not* the token contents.  For example, something that looks like `SQynup6xFEIshplTe-aj1mo-owqxx4n8SaiEgQvRa9E`.  And it is not the secret name.  For example, something that looks like `jenkins-token-m417e` (one of the secret names from my currently running jenkins service account).

I do not agree that multiple scoped tokens per user name is going to be a problem.  

Providing the token as input, the result of the self-sar requests  for the admin, edit, and view roles is what is important.  

Either whatever token provided has the admin, edit, or view role, or it does not.  Based on the role it has, the code maps it to one of the 3 correlating admin, edit or view permissions for the username our code will create in the Jenkins permission matrix.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: "userinfo" is essentially the "ping" ... it accesses the "/oapi/v1/users/~" URI on the master and sees if a user can read his own user info ... i.e. the user even exists.  If not, we don't either bother with the self-sar requests.

the "profile" object on the jenkins User object is just a means for updating metadata.  It provides access to any emails for example that the back end store (openshift oauth in this case) provides for the user.

The real key is these lines:

```
        UsernamePasswordAuthenticationToken token =
                new UsernamePasswordAuthenticationToken(matrixKey, "", authorities);
        SecurityContextHolder.getContext().setAuthentication(token);
```

But based on trying to explain this, I did find one flaw (I don't *think* it is exactly what you've been citing, but it is in the same ballparck).  I'll update the code with the fix I think it needs, plus some comments.  Hopefully that both shores things up and makes it clearer to observers to make sense of things.

gabemontero: Actually I think it is what you have been citing, at least close enough ... the appending "browser" vs. "token" I don't think is the way to go about addressing it.  I'll post a solution and a bit and we'll see if that still holds.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftPermissionFilter.java
gabemontero: Yes, "it's ok that the lookup returns an entry w/o a valid token".  cacheEntry.token can be null if the user supplied token previously failed the openshift oauth flows.  If so, and the time of the last check is still withing the polling interval, we'll immediate send the 401 without making another oauth req against openshift.  Then, when we've waited past the polling interval, we'll retry the openshift oauth reqs with the token supplied on the http req, and see if it is valid.

gabemontero:
OK ... i've pushed the update.  It covers the hole where there is a valid user "foo", and then a malicious user creates "foo-admin" without any of the admin/edit/view roles/permissions.  If "foo-admin" has only the "view" role, the prior logic, now encapsulated in the `if (suffix.length() !=0)` block still holds.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -605,97 +612,135 @@ public HttpResponse onSuccess(String authorizationCode) {
...
+                
+        // we append the role suffix to the name stored into Jenkins, since a given user is able to log in at varying scope/permission
+        // levels in openshift; however, for now, we make sure the display name for Jenkins does not include this suffix
+        String suffix = "";
+        if (allowedRoles.contains("admin"))
+            suffix = "-admin";
+        else if (allowedRoles.contains("edit"))
+            suffix = "-edit";
+        else if (allowedRoles.contains("view"))
+            suffix = "view";
```
bparees: should be "-view"?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -605,97 +612,135 @@ public HttpResponse onSuccess(String authorizationCode) {
...
-        	GlobalMatrixAuthorizationStrategy existingAuthMgr = (GlobalMatrixAuthorizationStrategy) Jenkins.getInstance().getAuthorizationStrategy();
-        	Set<String> usersGroups = existingAuthMgr.getGroups();
-            List<PermissionGroup> permissionGroups = new ArrayList<PermissionGroup>(PermissionGroup.getAll());
-            GlobalMatrixAuthorizationStrategy newAuthMgr = null;
-            if (existingAuthMgr instanceof ProjectMatrixAuthorizationStrategy) {
-            	newAuthMgr = new ProjectMatrixAuthorizationStrategy();
-            } else {
-            	newAuthMgr = new GlobalMatrixAuthorizationStrategy();
-            }
+        if (suffix.length() != 0) {
```
bparees: if we introduced a "-none" prefix that was appended by default, would that remove the need for this?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: -view is consistent with the others (though technically it works with just "view" ... i'll update and push

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: Yes a -none suffix would achieve similar affect.  I'll update to do it that way.

gabemontero:
Pushed the "-view" and "-none" changes.

gabemontero:
Dahhhhhhh ... wait, if someone had inconceivably created a valid user named "foo-none" with the admin role, and then the malicious user created "foo", his self sar would result in a "-none" suffix, and he'd get admin role access.

Moving back to the if suffix.length approach.

gabemontero:
Ok, still has "-view" correction, but backed off "-none"

bparees:
> wait, if someone had inconceivably created a valid user named "foo-none" with the admin role, and then the malicious user created "foo", his self sar would result in a "-none" suffix, and he'd get admin role access.

@gabemontero why?  the real "foo-none" user would be keyed on "foo-none-admin" in the matrix, it shouldn't apply to the malicious "foo" user.

bparees:
(this is all predicated on the assumption that the username is *always* suffixed for purposes of performing access/auth checks)

gabemontero:
Your are right (spun myself a bit dizzy on this one).  And yes, the username always being suffixed is required.

Thinking about it some more, I'd rather avoid interacting with Jenkins at all (i.e. not setting "foo-none" user on the security context, no interactions at all with the auth matrics, for a user who has same permissions as anonymous) that is achieved via `if (suffix != null)` vs. using "-none".

Based on that clarification, what is your opinion for this one then?

bparees:
> Based on that clarification, what is your opinion for this one then?

not sure.  i can see the argument for your point (if the default state is no permissions, don't do extra work), and for my point (consistently always add a suffix and enforce permissions via the suffix).

the latter feels a little more consistent+assertive to me and thus i think i have a slight preference for it.  but i'm also not sure i understand the whole flow+relationships between openshift users, tokens, jenkins users, and matrix auth entries as well as i need to to have an intelligent opinion.  maybe i'll grab you for 20 minutes tomorrow and have you whiteboard it out for me.

gabemontero:
OK yeah let's shoot for 15 to 20 of face time Wednesday to nail down the
last couple of bits here.

bparees:
lgtm

PR #13, opened 2017-05-12
move jenkins image content around login to this readme

gabemontero:
@bparees ptal

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,20 +1,84 @@
 openshift-login
 ===============
 
-A Jenkins plugin which lets you login to Jenkins with your account on your OpenShift installation. Focuses on enabling SSO for Jenkins within your cluster.
+A Jenkins plugin which lets you login to Jenkins with your account on an OpenShift installation.
 
 
-The primary scenario is using this plugin when Jenkins is running in a OpenShift pod, and the jobs with Jenkins operate against the
-same OpenShift cluster that Jenkins is running in.  In this scenario, no additional configuration is required, but you must be running agaist v1.4 of OpenShift/Origin (https://github.com/openshift/origin/tree/release-1.4).
+Primary scenario
+--------------------------------
 
-Otherwise, for development purposes, or for scenarios where the OpenShift pod defaults do not apply, the configuration parameters are:
+This plugin can function with no additional configuration within Jenkins, but you must be running in an OpenShift Pod and against v1.4+ of OpenShift/Origin (https://github.com/openshift/origin/tree/release-1.4).
 
-* service account directory:  The directory to load service account information from. Three files are referenced:  'namespace', 'ca.crt', and 'token'. They correspond to the OpenShift project, certificate, and authentication token for the service account.
-* service account name:  override for the service account name used when authenticating users against OAuth (default derived from token / client secret)
-* server prefix:  URI for the OpenShift OAuth endpoint
-* redirect URL: URL for the OpenShift API server
-* client ID:  override for the ID for the OpenShift OAuth client (default derived from service account information)
-* client secret:  override for the service account token (to change permissions for the OAuth client during the OAuth authentication flows)
+When running against a sufficient level of OpenShift/origin, and the plugin is installed in your Jenkins instance, the authentication mechanism (the "Security Realm") established within your Jenkins instance is as follows:
+
+* If running outside of an OpenShift Pod, then on start up the authentication mechanism configured for Jenkins is used.  
+* If running inside of an OpenShift Pod and the environment variable `OPENSHIFT_ENABLE_OAUTH` set to `false` on the container, then on start up the authentication mechanism configured for Jenkins is used. 
+* Otherwise, if running in an OpenShift Pod and the environment variable `OPENSHIFT_ENABLE_OAUTH` is set to a value other than `false` on the container, the plugin
+auto-enables itself to manage the login process, and to login you specify valid credentials as required by the identity provider used by OpenShift.  
+
+NOTE:  When this plugin manages authentication, the predefined `admin` user in the default Jenkins user database for the [OpenShift Jenkins image](https://github.com/openshift/jenkins) is now ignored.
+Unless there is an `admin` user defined within OpenShift with sufficient permissions to the project Jenkins is running in, you will not be able to do anything with Jenkins by logging in as `admin`.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,20 +1,82 @@
...
+* If running inside of an OpenShift Pod and the environment variable `OPENSHIFT_ENABLE_OAUTH` set to `false` on the container, default Jenkins authentication is again used, with the same specifics on the credentials to use from the prior bullet applying. 
+* Otherwise, if running in an OpenShift Pod and the environment variable `OPENSHIFT_ENABLE_OAUTH` is set to a value other than `false` on the container, the plugin
+auto-enables itself to manage the login process, and to login you specify valid credentials as required by the identity provider used by OpenShift.  In this case, the predefined `admin` user in the default Jenkins user database is now ignored.
+Unless there is an `admin` user defined within OpenShift with sufficient permissions to the project Jenkins is running in, you will not be able to do anything with Jenkins by logging in as `admin`.
+
+Running in an OpenShift Pod against v1.4 or later of OpenShift/Origin is hence the primary scenario for this plugin.
```
bparees: replace last sentence with: 

Running in an OpenShift Pod against v1.4 or later of OpenShift/Origin with `OPENSHIFT_ENABLED_OAUTH=true` is hence the primary scenario for this plugin.

README.md
```diff
@@ -1,20 +1,82 @@
...
-* server prefix:  URI for the OpenShift OAuth endpoint
-* redirect URL: URL for the OpenShift API server
-* client ID:  override for the ID for the OpenShift OAuth client (default derived from service account information)
-* client secret:  override for the service account token (to change permissions for the OAuth client during the OAuth authentication flows)
+When running against that level of OpenShift/origin, and the plugin installed in your Jenkins instance, the authentication mechanism (the "Security Realm") established within your Jenkins instance is as follows:
+
+* If running outside of an OpenShift Pod, then on start up default Jenkins authentication is used, irregardless of the setting of the environment variable `OPENSHIFT_ENABLE_OAUTH`.  If you are using the [OpenShift Jenkins image](https://github.com/openshift/jenkins), you log in with the user name `admin`, supplying the password specified by the `JENKINS_PASSWORD` environment variable set on the container. If you do not override `JENKINS_PASSWORD`, the default password for `admin` is `password`.  For any other pre-existing Jenkins installation, your credentials into that specific system are needed.
```
bparees: Is this first sentence true?  I thought the idea was "if running outside of openshift, and openshift_enable_oauth is true (and you've properly configured the plugin as described at the end of this doc?), you'll use openshift credentials to login, otherwise standard jenkins auth applies (ie this plugin does nothing)"

i'm also not sure we should talking about the JENKINS_PASSWORD variable here (or the default "admin" user for that matter) since that's behavior that's specific to the jenkins image, it's not behavior introduced/provided by this plugin.

gabemontero: `OPENSHIFT_ENABLE_OAUTH` only controls whether the plugin attempts to insert itself as the auth mechanism when it is not already configured as the mechanism

if you explicitly configure it as the mechanism in jenkins, you don't need to take the step of setting `OPENSHIFT_ENABLE_OAUTH`

On the `admin` ID and `JENKINS_PASSWORD`, I was unsure from our IRC exchange if you wanted that element moved as well, so I flipped a coin;  I am certainly comfortable with moving it back, with some sort of generic preamble that if you are not using login plugn, then ...

bparees: ok, in that case the statement is still a bit misleading then.  I know we're trying to avoid the gorpy details about configuring the plugin manually, but someone who reads that sentence may not even make it down to the gorpy details below, or be confused about what to do with the OPENSHIFT_ENABLE_OAUTH variable.

So i think it needs to be clarified either here, or in the secondary scenario section below.

bparees: > On the admin ID and JENKINS_PASSWORD, I was unsure from our IRC exchange if you wanted that element moved as well, so I flipped a coin; I am certainly comfortable with moving it back, with some sort of generic preamble that if you are not using login plugn, then

yeah, honestly I think the other doc probably says enough about it even after your edits (the JENKINS_PASSWORD variable is described) but this is what i was hinting at when i said i'd leave that PR open for you to merge in case you wanted to revisit it. :)

README.md
```diff
@@ -1,20 +1,82 @@
...
+A Jenkins plugin which lets you login to Jenkins with your account on an OpenShift installation.
 
 
-The primary scenario is using this plugin when Jenkins is running in a OpenShift pod, and the jobs with Jenkins operate against the
-same OpenShift cluster that Jenkins is running in.  In this scenario, no additional configuration is required, but you must be running agaist v1.4 of OpenShift/Origin (https://github.com/openshift/origin/tree/release-1.4).
+Primary scenario
+--------------------------------
 
-Otherwise, for development purposes, or for scenarios where the OpenShift pod defaults do not apply, the configuration parameters are:
+This plugin can function with no additional configuration within Jenkins, but you must be running in an OpenShift Pod and against v1.4 of OpenShift/Origin (https://github.com/openshift/origin/tree/release-1.4).
```
gabemontero: part of next push

README.md
```diff
@@ -1,20 +1,82 @@
...
 
-* service account directory:  The directory to load service account information from. Three files are referenced:  'namespace', 'ca.crt', and 'token'. They correspond to the OpenShift project, certificate, and authentication token for the service account.
-* service account name:  override for the service account name used when authenticating users against OAuth (default derived from token / client secret)
-* server prefix:  URI for the OpenShift OAuth endpoint
-* redirect URL: URL for the OpenShift API server
-* client ID:  override for the ID for the OpenShift OAuth client (default derived from service account information)
-* client secret:  override for the service account token (to change permissions for the OAuth client during the OAuth authentication flows)
+When running against that level of OpenShift/origin, and the plugin installed in your Jenkins instance, the authentication mechanism (the "Security Realm") established within your Jenkins instance is as follows:
```
gabemontero: part of next push

README.md
gabemontero: entire line reworked

README.md
gabemontero: part of next push

gabemontero:
ok updates pushed

bparees:
lgtm

PR #30, opened 2018-01-08
make sure filter inserted after restarts on PVs

gabemontero:
@openshift/sig-developer-experience @stevekuznetsov fyi / ptal

this fixes @stevekuznetsov 's can't use token without browser login after restarting jenkins in a pod with PVs problem


gabemontero:
[test]

bparees:
nice

PR #31, opened 2018-02-01
Log token information in debug instead of info

chmouel:
Too much things is logged there (i.e: token), let's get that in
debug instead.

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -423,8 +423,10 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
                             this.defaultedServiceAccountDirectory,
                             this.serviceAccountName,
                             this.defaultedServiceAccountName, this.clientId,
-                            this.defaultedClientId, this.clientSecret,
-                            this.defaultedClientSecret, this.redirectURL,
+                            this.defaultedClientId,
+                            Secret.toString(this.clientSecret).substring(0, 5) + ".......",
+                            this.defaultedClientSecret.substring(0, 5) + ".......",
+                            this.redirectURL,
                             this.defaultedRedirectURL, this.serverPrefix,
                             this.defaultedServerPrefix));
 


gabemontero:
Thanks for the submission @chmouel 

This was actually an area of focus during the initial delivery of this plugin, and a bit of a non-standard situation IMO.  In fact there were times in the past when it was not an info level message.  

The difficulty stemmed from debugging user issues when things did not function (typically because of user misconfiguration).

They could not log into Jenkins to bump the logging level to debug, etc. so we could get the information, nor had the expertise, etc. to change the logging level via direct manipulation of the config files.

Hence making my job much more difficult :-)

I might be amenable to say obfuscating the token by default, or changing the output to simply print where the token came from (default pod location vs. explicit configuration, for example), maybe see about condensing some of the other pieces.  If you are willing to work this PR further, let's see what that would look like.

If you don't have that amount of bandwidth, I'm good with opening an issue to track doing just that, where I'll work on it as time permits (I'll probably have some time next week if that helps your decision making).

But I'm not making a blanket info to debug conversion like this.

Thanks again ... let me know what you think.

chmouel:
Thanks for the detailed info @gabemontero , I have updated this commit. My Java is humm rusty so if you feel like this could better done feel free to trash this and implement it properly! 🔨

gabemontero:
NP /YW @chmouel 

re: Java particulars looks fine to me

You happen to have some sample output handy?  If so, post here, and assuming it looks good I'll merge.

If not I'll pull the change and try it out myself, just let me know.

gabemontero:
[test]

chmouel:
re: You happen to have some sample output handy? If so, post here, and assuming it looks good I'll merge.

unfortunately not for the new format, testing this in prod involve quite a bit of hoopla hoops and I would assume this is straightforward enough to get trusted by code review and jenkins tests,

gabemontero:
cool - no problem

i should be able to check out your branch and give it a whirl either later today or first thing Friday.  There is also a chance our PR test might show it (don't remember if it does off the top of my head).

I'll report back when I have the data.  Thanks again.

gabemontero:
Yep test job had the data ... but looking at it, I don't think the change had the impact you wanted @chmouel: 

```
INFO: OpenShift OAuth: provider: OpenShiftProviderInfo: issuer: https://ec2-34-238-158-252.compute-1.amazonaws.com:8443 auth ep: https://ec2-34-238-158-252.compute-1.amazonaws.com:8443/oauth/authorize token ep: https.......
Feb 01, 2018 7:31:51 PM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm populateDefaults
INFO: OpenShift OAuth returning true with namespace extended-test-jenkins-plugin-md5sd-rf54s-jenkins SA dir null default /run/secrets/kubernetes.io/serviceaccount SA name null default jenkins client ID null default system:serviceaccount:extended-test-jenkins-plugin-md5sd-rf54s-jenkins:jenkins secret null default eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJleHRlbmRlZC10ZXN0LWplbmtpbnMtcGx1Z2luLW1kNXNkLXJmNTRzLWplbmtpbnMiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlY3JldC5uYW1lIjoiamVua2lucy10b2tlbi1zYmdzYiIsImt1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvc2VydmljZS1hY2NvdW50Lm5hbWUiOiJqZW5raW5zIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZXJ2aWNlLWFjY291bnQudWlkIjoiNmM5ZDQ1MjUtMDc4Ni0xMWU4LTk5NjktMGVjZGEwNjVhMzQ4Iiwic3ViIjoic3lzdGVtOnNlcnZpY2VhY2NvdW50OmV4dGVuZGVkLXRlc3QtamVua2lucy1wbHVnaW4tbWQ1c2QtcmY1NHMtamVua2luczpqZW5raW5zIn0.C_TyMRiQab_SYYP1ICQ-Z2BFFs17-66s8bJ7bZiCQIxuAmR7H-1AInZexo0jDtmHbxafjRjchoSUnMgxNUNX0DD4Uavd9olgdxdiQ3Z8N_mBcFSeV4CQoSvqmXz5sP3dVG_jWdaTcHcaobQLOSyJUaECSskBMmJU_uxA5CxtODiwevpO3tkNpWYOldWQiXze6EQAVQVOxNVvLVoLLvLemzhZo6o-wivlEpimELa4P08zsuh-BxDFX0C5s1i-yCb0x389IoBVvRYI1RYtqOnCtTKezvIIct_sIAbhYcg9lFJ6cuPRBnPh_k738DykXuSja8pLJcNDv3VM8EXM6ssqDg redirect null default https://ec2-34-238-158-252.compute-1.amazonaws.com:8443 server null default https://openshift.default.svc

```
The change ended up obfuscating the endpoint address.

Upon further review, I believe we should remove what you did at OpenShiftProviderInfo and leveage the same technique at https://github.com/openshift/jenkins-openshift-login-plugin/blob/master/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java#L427

chmouel:
sounds right, thanks, i'll amend that by tomorrow!

chmouel:
[test]

chmouel:
I think that's about it, I removed my old commits and pushed/forced. Hopefully i haven't missed something,

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -423,8 +423,10 @@ else if (withinAPod)
                             this.defaultedServiceAccountDirectory,
                             this.serviceAccountName,
                             this.defaultedServiceAccountName, this.clientId,
-                            this.defaultedClientId, this.clientSecret,
-                            this.defaultedClientSecret, this.redirectURL,
+                            this.defaultedClientId,
+                            this.clientSecret.substring(0, 5) + ".......",
+                            this.defaultedClientSecret(0, 5) + ".......",
+                            this.redirectURL,
```
gabemontero: You got a compile error:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /data/src/github.com/openshift/jenkins-openshift-login-plugin/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java:[427,45] error: cannot find symbol
[ERROR]   symbol:   method substring(int,int)
  location: variable clientSecret of type Secret
/data/src/github.com/openshift/jenkins-openshift-login-plugin/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java:[428,32] error: cannot find symbol
[INFO] 2 errors 

```

`clientSecret` is of type hudson.util.Secret

Use `Secret.toString(this.clientSecret).substring(0, 5) + "......."`

it accounts for null, but doesn't further encode (to again make diagnostics more difficult)

chmouel: Silly me, and me thinking it was a quick fix, one should never underestimate a quick fix when  not skilled in the say language :) i'll use a proper ide and send an update, thanks for your patience again.

chmouel:
[test]

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: no problem and thanks for the effort

gabemontero:
Test failed because of 

```
WARNING: null
java.lang.StringIndexOutOfBoundsException: String index out of range: 5
	at java.lang.String.substring(String.java:1963)
	at org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm.populateDefaults(OpenShiftOAuth2SecurityRealm.java:427)
	at org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm.<init>(OpenShiftOAuth2SecurityRealm.java:281)
	at org.openshift.jenkins.plugins.openshiftlogin.OpenShiftSetOAuth.setOauth(OpenShiftSetOAuth.java:74)
	at org.openshift.jenkins.plugins.openshiftlogin.OpenShiftPermissionFilter.doFilter(OpenShiftPermissionFilter.java:122)
```

We've got to test the length of the string before calling substring ... it is possible it could be null.

Let me know @chmouel if this is taking more time for you than it is worth for you, and you'd rather have me code this up in a separate PR.

thanks

gabemontero:
Yeah I ended having a second and have it working @chmouel .

I've got a change that tweaks what you did and produces:

```
Feb 02, 2018 6:40:41 PM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm populateDefaults
INFO: OpenShift OAuth returning true with namespace myproject SA dir null default /run/secrets/kubernetes.io/serviceaccount SA name null default jenkins client ID null default system:serviceaccount:myproject:jenkins secret null default eyJhb....... redirect null default https://127.0.0.1:8443 server null default https://openshift.default.svc
```

I'll go ahead and submit my PR and reference it here.

PR #33, opened 2018-02-19
Add user to usersGroup explicitly to avoid edge case

steven-terrana:
When configuring this plugin via a groovy init script running during Jenkins initialization it was discovered that user's roles will not map correctly if there are no existing user groups defined in the authorization strategy, as would be possible via:

~~~ groovy
GlobalMatrixAuthorizationStrategy newAuthMgr = new GlobalMatrixAuthorizationStrategy()
Jenkins.getInstance.setAuthorizationStrategy(newAuthMgr)
~~~

By immediately adding the matrixKey to the usersGroups Set, there will always be at least 1 element for the `for (String userGroup : usersGroups)` loop to iterate over, thus mapping the roles appropriately in this situation.

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -784,7 +784,7 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
                             .format("OpenShift OAuth: user %s, stored in the matrix as %s, based on OpenShift roles %s already exists in Jenkins",
                                     info.getName(), matrixKey, allowedRoles));
                 } else {
-                    userGroups.add(matrixKey);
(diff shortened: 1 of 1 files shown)


gabemontero:
[test]

gabemontero:
Thanks for the contribution @steven-terrana !  Aside from notifying us of the issue, the PR obviously jump started resolving this.

That said, after studying it a bit, I think rather than adding the matrixKey to the user group set returned from Jenkins, we should simply move our view/edit/admin process for the new user out of the `userGroups` for loop.  Among other things
- if jenkins starts returning a copy of that Set, our update will be lost
- upon reflection, we are unnecessarily adding the users view/edit/admin permssions multiple times ... for each user :-) .... we only need to do it once.

Unless you are still online and want to update your PR, great.  But I'll start crafting a new PR in the interim and get it through if I don't see updates from you before I finish.

thanks again

gabemontero:
Closing this in favor of https://github.com/openshift/jenkins-openshift-login-plugin/pull/34

PR #38, opened 2018-03-19
more adjusting for v2.107 changes

gabemontero:
@openshift/sig-developer-experience ptal

comments in the code get into the details ... but some more changes as a result of testing on v2.107 of jenkins


gabemontero:
@stevekuznetsov - second instance now with this repro where the prow commands are not working for us.

the owners file looks right

not sure where to look to see what is wrong ... can you take a look / provide us a pointer ?

PR #39, opened 2018-03-28
one more spot of explicit exception logging

gabemontero:
@openshift/sig-developer-experience fyi

adding this log in a supplied patch assisted in diagnosing a customer issue (where the certs from the master/node were bad) recently discussed on openshift-sme

getting the customer to turn on trace and restart jenkins was problematic ... in theory, the logging will only occur if they are configured to do oauth and want to know what is going on, but added some rudimentary throttling in case somebody has a script with a for loop that is continually hitting the jenkins endpoint ... or our liveness check is tuned super fast, as the filter chain will be executed


gabemontero:
`Stage FORWARD GCS CREDENTIALS TO REMOTE HOST [00h 00m 00s] failed.`

gabemontero:
this time it was some hiccup with the git processing when building origin images 

assuming it is a flake for now

gabemontero:
hiccup with install origin

/retest

gabemontero:
PR testing here is blocked by https://github.com/openshift/origin/issues/19143 ... I'll manually test/merge if it persists

gabemontero:
an AMI update has occurred which should fix things ... let's see

/retest

PR #42, opened 2018-06-20
Adding job cancel permission for the edit role

### LINKED ISSUE #41: Edit role provide cancel job permission
Right now, the permissions are very fixed for what "admin", "edit", and "view" can provide. Would there be a way to customize the permissions matrix for users in these specific roles?

I would like to give our developers Item.CANCEL permission, and in the namespaces that the Jenkins containers are running, the developers have the edit role. Is there any way that I could do this without writing my own update to the code and build the plugin separately?

jmcshane:
Resolves #41


gabemontero:
@openshift/sig-developer-experience fyi

PR #43, opened 2018-07-18
Enhanced non-browser access documentation on how to leverage token from OpenShift, cleaned up code blocks

andyfeller:


andyfeller:
@gabemontero : here's to you

Issue #44, opened 2018-10-02
Jenkins blocks edit role's Jenkins.RUN_SCRIPTS permission without Jenkins.ADMINISTER permission due to matrix-auth-plugin


andyfeller:
[org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm. updateAuthorizationStrategy()](https://github.com/openshift/jenkins-openshift-login-plugin/blob/master/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java#L840) grants those with `edit` OpenShift role the `Jenkins.RUN_SCRIPTS` permission, which I was banking on, however the [jenkinsci/matrix-auth-plugin](https://github.com/jenkinsci/matrix-auth-plugin/blob/master/src/main/resources/hudson/security/DangerousMatrixPermissionsAdministrativeMonitor/message.properties) denies this permission without `Jenkins.ADMINISTER`.  So I'm a little torn because I want the ability to run scripts for edit purposes because my system administrators won't give us `admin` role for service accounts and we need to configure Jenkins in a non-static way.

Thoughts?

gabemontero:
Ooh ... this looks like a recent change in the matrix auth plugin that went out with their 2.3 version.

Our image for example currently pulls in 2.2.  But if we had to upgrade for say a jenkins advisory (a quite likely scenario) we will hit issues running scripts.

I'm leaning towards updating our mapping to add `Jenkins.ADMINISTER` to our edit role, as being able to run scripts is key, sensible mapping between administrator notions be damned.

@bparees opinions?

andyfeller:
Thanks for the follow up @gabemontero! 🙇 

Yeah, we were thinking about adding the `Jenkins.ADMINISTER` permission for OpenShift `edit` role, too; otherwise we'd need a way to override the name of the OpenShift role to translate to Jenkins auth matrix.

bparees:
What distinguishes someone with openshift admin from someone with openshift edit at that point, in terms of their jenkins permissions?

andyfeller:
@bparees : think that is a fair question.  one possible response is that nothing distinguishes them as Overall/Administer is enough to do everything assigned to those with OpenShift `admin` role, so why not make them one in the same.  In this case, then I think the openshift-login plugin should allow someone to decide which OpenShift role(s) match to which Jenkins permissions.

gabemontero:
[...]

PR #45, opened 2018-10-12
make the list of openshift roles and the role to jenkins permission m…

gabemontero:
…apping configurable

See https://github.com/openshift/jenkins-openshift-login-plugin/issues/44
See https://trello.com/c/kDqWZlOH/1207-5-allow-the-openshift-role-to-jenkins-permission-mapping-to-be-configurable-jenkinsintegration

amongst others

@openshift/sig-developer-experience ptal


gabemontero:
On Sat, Oct 13, 2018 at 11:26 PM Ben Parees <notifications@github.com>
wrote:

> *@bparees* requested changes on this pull request.
>
> mostly looks good, main thing i'd like to see is a model that takes all of
> a user's roles into account instead of just the "top ranked" one because i
> can foresee admins wanting to setup disjoint/orthogonal roles that don't
> have a relative ranking and/or that some users will have both roles.
>
 You can still have multiple roles available.  It is just picking which
role we do the first self SAR with.  If they had disjoint/orthogonal roles,
each of those roles would have a separate token or form of auth, no?

If so, they supply the token/auth for the role they want for that session.
I forgot the name of the feature that enabled that ... it was the thing we
discussed with Jordan when we set up the current approach.

> Admittedly they could introduce a new role that encompassed the other two
> roles, rank it higher, and give it to those users, but that seems like an
> annoying workaround.
> ------------------------------
>
> In README.md
> <https://github.com/openshift/jenkins-openshift-login-plugin/pull/45#discussion_r224977506>
> :
>
> > @@ -91,7 +91,25 @@ For the `edit` role, in addition to the permissions available to `view`:
>
>  Users authenticated against OpenShift OAuth will be added to the Jenkins authorization matrix upon their first successful login.
>
> -Permissions for users in Jenkins can be changed in OpenShift after those users are initially established in Jenkins.  The OpenShift Login plugin polls the OpenShift API server for permissions and will update the permissions stored in
> +Now, with v1.0.10 of this plugin, you can change
> +
> +* Which OpenShift Roles are checked for authorization
> +* Which Jenkins permissions map to which OpenShift Roles
> +
> +This plugin will look at for a ConfigMap named openshift-jenkins-login-plugin-config.  If it finds and can read in that ConfigMap, it then:
>
> looks for it in the same namespace as jenkins is running in, presumably
>
typically yes though precisely it is the namespace housing the service
account that acts initiates the oauth proxy flow.  The one annotated on the
openshift side etc etc.   Technically it could be in any namespace.  And
technically rhus stuff can work if jenkins is running outside of openshift

> ------------------------------
>
> In README.md
> <https://github.com/openshift/jenkins-openshift-login-plugin/pull/45#discussion_r224977514>
> :
>
> > @@ -91,7 +91,25 @@ For the `edit` role, in addition to the permissions available to `view`:
>
>  Users authenticated against OpenShift OAuth will be added to the Jenkins authorization matrix upon their first successful login.
>
> -Permissions for users in Jenkins can be changed in OpenShift after those users are initially established in Jenkins.  The OpenShift Login plugin polls the OpenShift API server for permissions and will update the permissions stored in
> +Now, with v1.0.10 of this plugin, you can change
> +
> +* Which OpenShift Roles are checked for authorization
> +* Which Jenkins permissions map to which OpenShift Roles
> +
> +This plugin will look at for a ConfigMap named openshift-jenkins-login-plugin-config.  If it finds and can read in that ConfigMap, it then:
> +
> +* Checks the `data` map for a key named `roleRanking`
> +* If found, that defines the list of OpenShift Roles to perform authorization review against.
>
> comma separated list? space separated?
>
comma separated but yeah I’ll update that to make it clear

> ------------------------------
>
> In README.md
> <https://github.com/openshift/jenkins-openshift-login-plugin/pull/45#discussion_r224977549>
> :
>
> > +
> +* Which OpenShift Roles are checked for authorization
> +* Which Jenkins permissions map to which OpenShift Roles
> +
> +This plugin will look at for a ConfigMap named openshift-jenkins-login-plugin-config.  If it finds and can read in that ConfigMap, it then:
> +
> +* Checks the `data` map for a key named `roleRanking`
> +* If found, that defines the list of OpenShift Roles to perform authorization review against.
> +* The order of the list matters. The first entry should have the most permissions in Jenkins. So it is the equivalent of the `admin` role with our default settings.
> +* The authorization reviews are considered complete with the first successful result.  So, just like with our defaults, the most authorized role is used.
> +* Any other key/value pair is consider a Jenkins permission to OpenShift Role mapping.
> +* The key is the Jenkins permission group short ID and the Jenkins permission short ID, with those two separated by a hyphen or `-` character.
> +* So if you want to add the Overall Jenkins Administer permission to an OpenShift Role, the key should be `Overall-Administer`
> +* To get a sense of which permission groups and permissions IDs are available, go the the matrix authorization page in the Jenkins console and IDs for the groups and individual permissions in the table they provide.
> +* The value of the key/value pair is the list of OpenShift Roles the permission should apply to, with each role separated by a comma or `,`.
> +* So if you want to add the Overall Jenkins Administer permission to say both the default `admin` and `edit` roles, as well as a new `jenkins` role you have created, the value for the key `Overall-Administer` would be `admin,edit,jenkins`.
>
> does this order matter too?
>
> alternatively what happens if someone lists a role here that's not in
> roleRanking? (i guess users with that role would never be matched to this
> permission since we'd never check/confirm they had that permission)
>
> I think the mental model here is "1) walk the list of roleRankings, stop
> on the first success. 2) Scan the various permission keys and see if any of
> them include the role found in (1) in their value, if so, grant that
> permission to the user"?
>
yea

> It might be useful to describe it that way if that's accurate, i think the
> current description risks a disconnect between what roleRanking does and
> how the permission key/value pairs work.
>
ok I’ll send an update to hopefully clear it up

> I also wonder if it would be better to check *all* the roles, rather than
> stopping on the first one. That way you can set things up for disjoint
> roles like:
>
> roles: role1,role2,role3
> SomePermForRole1: role1
> SomePermForRole2: role2
>
> well that is not how it works today

>  which under the current scheme (if i've understood it correctly), would
> fail for a user who has both "role1" and "role2" (they wouldn't get the
> SomePerForRole2 permission granted, because we'd have stopped once we saw
> they had role1), and the "fix" would be to list both role1+role2 under
> SomePermForRole2 which might not be desirable (not everyone w/ role1 should
> get that permission)
>
yeah per my earlier comment that does not line up with the understanding of
how the auth works

Maybe we just in front of a white board and has it out

> ------------------------------
>
> In
> src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
> <https://github.com/openshift/jenkins-openshift-login-plugin/pull/45#discussion_r224977651>
> :
>
> > +        permMap.put("admin", adminPerms);
> +
> +        final Credential credential = new Credential(
> +                BearerToken.authorizationHeaderAccessMethod())
> +                .setAccessToken(getDefaultedClientSecret().getPlainText());
> +        HttpRequestFactory requestFactory = transport
> +                .createRequestFactory(new HttpRequestInitializer() {
> +                    public void initialize(HttpRequest request)
> +                            throws IOException {
> +                        credential.initialize(request);
> +                        request.setParser(new JsonObjectParser(JSON_FACTORY));
> +                    }
> +                });
> +        GenericUrl url = new GenericUrl(getDefaultedServerPrefix() + String.format(CONFIG_MAP_URI, namespace));
> +        HttpRequest request = null;//
> +        ConfigMapResponse response = null;//
>
> trailing //s ?
> ------------------------------
>
> In
> src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
> <https://github.com/openshift/jenkins-openshift-login-plugin/pull/45#discussion_r224977726>
> :
>
> > @@ -614,10 +617,144 @@ public void initialize(HttpRequest request)
>                                      Boolean.toString(review.allowed),
>                                      review.reason));
>                  if (review.allowed)
> -                    allowedVerbs.add(verb);
> +                    return verb;
> +            }
> +        }
> +        return null;
> +    }
> +
> +    private Map<String, List<Permission>> getRoleToPermissionMap(final HttpTransport transport) {
>
> how often is this called? on any new login and on the periodic
> sync/refresh of the permissions?
>
it is part of the existing refresh permissions path

A refresher on that
- driven off of reqs for the user against the jenkins console
- last refresh time stored in the http session
- when the configured time between refresh has elapsed we do new refresh

bparees:
> If so, they supply the token/auth for the role they want for that session.

how do they do that when they're just providing their username+password to login to the console?

I thought the token they supply to jenkins(or the permissions they grant to jenkins via the oauth flow) is one that allows jenkins to perform sar checks against their account(not against a specific scoped token).  Not one that provides specific roles like "admin" or "view".  And then it's up to jenkins to perform the sar checks it wants to, and grant whatever jenkins permissions it wants to based on the results of the sar checks.

>  Maybe we just in front of a white board and has it out

yeah let's plan to talk through this during the arch call on monday, i certainly only have a passing understanding of how the oauth flow works currently.

README.md
```diff
@@ -91,7 +91,25 @@ For the `edit` role, in addition to the permissions available to `view`:
...
+* The authorization reviews are considered complete with the first successful result.  So, just like with our defaults, the most authorized role is used.
+* Any other key/value pair is consider a Jenkins permission to OpenShift Role mapping.
+* The key is the Jenkins permission group short ID and the Jenkins permission short ID, with those two separated by a hyphen or `-` character.
+* So if you want to add the Overall Jenkins Administer permission to an OpenShift Role, the key should be `Overall-Administer`
+* To get a sense of which permission groups and permissions IDs are available, go the the matrix authorization page in the Jenkins console and IDs for the groups and individual permissions in the table they provide.
+* The value of the key/value pair is the list of OpenShift Roles the permission should apply to, with each role separated by a comma or `,`.
+* So if you want to add the Overall Jenkins Administer permission to say both the default `admin` and `edit` roles, as well as a new `jenkins` role you have created, the value for the key `Overall-Administer` would be `admin,edit,jenkins`.
```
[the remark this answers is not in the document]
andyfeller: > I also wonder if it would be better to check all the roles, rather than stopping on the first one.

I'm having difficulty putting my finger on a specific doc saying it but I'm under the impression that OpenShift honors the aggregate of role bindings and would expect this to do the same as long as the permissions granted are additive.  We should ensure there is no inconsistency with permission models else there will be confusion and problems.

zonArt:
Thanks for this PR, it looks good, however I'm not sure to fully understand how it's suppose to work.
I've built the plugin from the source and am trying to play with it since it might answer to concerns I have. Unfortunately, I'm not able to make it work the way I understand it should. I guess you should be able to assign as many jenkins permissions you need to a specific role but when I try, I've only the last one assigned to my user. Let say I have the following configMap:

```
apiVersion: v1
data:
  Credentials-Create: view
  Credentials-View: view
  Job-Build: view
  Job-Cancel: view
  Job-Configure: view
  Job-ConfigureVersions: view
  Job-Create: view
  Job-Delete: view
  Job-Discover: view
  Job-Move: view
  Job-Read: view
  Job-Workspace: view
  Overall-Read: view
  Run-Replay: view
  SCM-Tag: view
  roleRanking: admin,edit,view
kind: ConfigMap
```
When I log in with a user which has the view role, I've only SCM-Tag permission added to my user lacking even the Overall-Read permission, is my configMap set the wrong way?

Second thing I noticed is that if I don't set the configMap in the namespace, I got a 404 error in jenkins logs telling me it doesn't find the expected configMap, then I'm unable to login. I don't have a log message telling that the configMap wasn't found and that default values will be used instead as mentioned in the code [here](https://github.com/openshift/jenkins-openshift-login-plugin/pull/45/files#diff-d68a65b0daea9ede707036e33ba65f4aR688)

gabemontero:
OK I've pushed the update to from a ranked role list to a generic role list, where I union the the permission set for all the roles the given user passes authorization for.  

Also includes the README clarifications outside of the previously asked for.  And I've changed `roleRanking` to `roleList`

@bparees @andyfeller ptal when you get the chance

@zonArt - per your comment
- I just re-verified with my version and if I don't have the config map, I just get the default perm mappings and I can log in
- your config map looks ok ... since you were using the default role names of admin, edit, and view ... you did not need to specify the role list, but it should not have hurt
- yeah, if you did not get the log you noted on the bad config map, it makes me wonder about the building of the plugin ... or some incompatibility with your specific jenkins that is not obvious to me
- try reproducing with the new commit I just pushed, and provide the complete jenkins logs (i.e. if you are running jenkins in a pod, dump the pod logs) ... and I can take a look

zonArt:
@gabemontero Thank you very much, I'll give it a try tomorrow and let you know. My version of jenkins is not the very latest (2.89.4 if I remember correctly, not at work right now) but I'm pretty sure it shouldn't matter. If I still experience issue, I'll dump the logs for you to help me figure out what's wrong

zonArt:
Hi,

I'll try against the latest commit and still got some issues, here is the stacktrace when configmap doesn't exist (https://gist.github.com/zonArt/bad9ecc0dbc546864a1ff10c0bedee4c)

My jenkins logs:
```
Oct 19, 2018 11:14:17 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin could not find the openshift-jenkins-login-plugin-config config map in namespace jenkins-docker so the default permission mapping will be used
Oct 19, 2018 11:15:41 AM org.eclipse.jetty.server.session.Session beginInvalidate
INFO: Session node0xwihjtajftadoz8pphdq86zt2 already being invalidated
Oct 19, 2018 11:15:41 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm populateDefaults
INFO: OpenShift OAuth: provider: OpenShiftProviderInfo: issuer: https://openshift.example.com:8443
```

zonArt:
Well, now the issue with unexisting configMap seems to have disappeared and if I connect inside jenkins' pod I can see that the permissions are added the right way (with configMap) so this looks good, however I'm not able to log in anymore (cf gist above with stacktrace), any clue?

README.md
```diff
@@ -91,7 +91,28 @@ For the `edit` role, in addition to the permissions available to `view`:
...
+
+This plugin will look at for a ConfigMap named openshift-jenkins-login-plugin-config.  Typically, when running Jenkins in an OpenShift Pod, it 
+will look in the namespace that Jenkins is running in.  Otherwise, it looks in the namespace specified in the "client ID" as explained in ["Secondary Scenarios" down below.](#secondary-scenarios) 
+
+If this plugin finds and can read in that ConfigMap, it then:
+
+* Checks the `data` map for a key named `roleList`
+* If found, the plugin expects it to be a comma separated list of OpenShift Roles to perform authorization review against.
+* The order of the list does not matter. The plugin will create a union set of the permissions associated with each Role that passes the authorization review.
+* The authorization reviews are considered complete with the first successful result.  So, just like with our defaults, the most authorized role is used.
```
bparees: doesn't this contradict the previous  bullet?

README.md
```diff
@@ -91,7 +91,28 @@ For the `edit` role, in addition to the permissions available to `view`:
...
+* Which OpenShift Roles are checked for authorization
+* Which Jenkins permissions map to which OpenShift Roles
+
+This plugin will look at for a ConfigMap named openshift-jenkins-login-plugin-config.  Typically, when running Jenkins in an OpenShift Pod, it 
+will look in the namespace that Jenkins is running in.  Otherwise, it looks in the namespace specified in the "client ID" as explained in ["Secondary Scenarios" down below.](#secondary-scenarios) 
+
+If this plugin finds and can read in that ConfigMap, it then:
+
+* Checks the `data` map for a key named `roleList`
+* If found, the plugin expects it to be a comma separated list of OpenShift Roles to perform authorization review against.
```
bparees: given that we know the list of roles they care about, from the roles they reference in their permission/role mapping entries, do we need this?  Can't we just construct the set of roles to SAR ourselves by looking at all the roles that are referenced in their mappings?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -613,11 +617,149 @@ public void initialize(HttpRequest request)
...
+        List<Permission> permissionsInSystem = Permission.getAll();
+        for (Entry<String,String> entry : response.data.entrySet()) {
+            String permStr = entry.getKey();
+            
+            if (permStr.trim().equalsIgnoreCase(ConfigMapResponse.ROLE_RANKING)) {
+                String roleList = entry.getValue();
+                if (roleList != null) {
+                    String[] newRoles = roleList.split(",");
+                    if (newRoles != null && newRoles.length > 0) {
+                        roles = new ArrayList<String>(Arrays.asList(newRoles));
```
bparees: per my comment above, i think we can just build this up by looking at all the roles we see when iterating the mappings.  Once less thing for users to have to specify and potentially screw up.

zonArt:
With the latest jenkins LTS (2.138.2) and this commit for the plugin, I'm now again stuck at not having all the needed permission, only the last one in alphabetical order, here is the jenkins logs:
```
INFO: OpenShift OAuth returning true with namespace example-namespace SA dir null default /run/secrets/kubernetes.io/serviceaccount SA name null default jenkins client ID null default system:serviceaccount:example-namespace:jenkins secret null default eyJhb....... redirect null default https://openshift.example.com:8443  server null default https://openshift.default.svc
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm populateDefaults
INFO: OpenShift OAuth: provider: OpenShiftProviderInfo: issuer: https://openshift.example.com:8443  auth ep: https://openshift.example.com:8443/oauth/authorize  token ep: https://openshift.example.com:8443/oauth/token 
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm populateDefaults
INFO: OpenShift OAuth returning true with namespace example-namespace SA dir null default /run/secrets/kubernetes.io/serviceaccount SA name null default jenkins client ID null default system:serviceaccount:example-namespace:jenkins secret null default eyJhb....... redirect null default https://openshift.example.com:8443  server null default https://openshift.default.svc
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Credentials-Create to Jenkins permission Permission[class com.cloudbees.plugins.credentials.CredentialsProvider,Create]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Credentials-Create for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Credentials-View to Jenkins permission Permission[class com.cloudbees.plugins.credentials.CredentialsProvider,View]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Credentials-View for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Build to Jenkins permission Permission[interface hudson.model.Item,Build]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Build for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Cancel to Jenkins permission Permission[interface hudson.model.Item,Cancel]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Cancel for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Configure to Jenkins permission Permission[interface hudson.model.Item,Configure]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Configure for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
WARNING: OpenShift Jenkins Login Plugin could not find permission Job-ConfigureVersions in Jenkins list of all available permissions
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Create to Jenkins permission Permission[interface hudson.model.Item,Create]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Create for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Delete to Jenkins permission Permission[interface hudson.model.Item,Delete]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Delete for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Discover to Jenkins permission Permission[interface hudson.model.Item,Discover]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Discover for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Move to Jenkins permission Permission[interface hudson.model.Item,Move]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Move for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Read to Jenkins permission Permission[interface hudson.model.Item,Read]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Read for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Job-Workspace to Jenkins permission Permission[interface hudson.model.Item,Workspace]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Job-Workspace for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Overall-Read to Jenkins permission Permission[class hudson.model.Hudson,Read]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Overall-Read for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission Run-Replay to Jenkins permission Permission[class hudson.model.Run,Replay]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission Run-Replay for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin matching configured permission SCM-Tag to Jenkins permission Permission[class hudson.scm.SCM,Tag]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin adding permission SCM-Tag for role view
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm getRoleToPermissionMap
INFO: OpenShift Jenkins Login Plugin using role list [admin, edit, view]
Oct 23, 2018 10:55:12 AM org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm updateAuthorizationStrategy
INFO: OpenShift OAuth: adding permissions for user ngy, stored in the matrix as ngy-view, based on OpenShift roles [view]
```

And here is a screenshot of what the matrix permission looks like in jenkins:
image

andyfeller:
Realize the _"Purple House"_ news is still fresh and likely things are somewhat chaotic, but wanted to nudge this PR along given @zonArt feedback.

gabemontero:
If some current higher priority threads complete as planned I hope to start spending time on this again beginning of next

gabemontero:
OK I've reproduced @zonArt 's problem and believe I have fixed it ... a combination of bugs were his config was subtly different than what I was testing with

I've pushed to update for you all to try

I'll circle back to @bparees 's last set of comments on tomorrow

gabemontero:
good news @zonArt !!

thanks for the confirmation, and the pre-release testing efforts :-)

README.md
gabemontero: it does ... forgot to remove this and we changed the flow ... removed in next push

README.md
gabemontero: agreed

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: agreed

gabemontero:
ok I've pushed the updates for @bparees 's suggestion, including removing the need for roleRanking/roleList 

if I can get our ci working again today or tomorrow we'll merge this the old fashion way, otherwise we'll green button this one.

gabemontero:
the ci fix worked and the build/test of the login plugin is occurring as I type ... it it passes, this will merge and I'll cut a new version of the plugin

gabemontero:
@piyush-garg @hrishin when this merges and I cut a new version of the plugin you should be good wrt what you all asked for in https://github.com/openshift/jenkins-openshift-login-plugin/issues/48

gabemontero:
v1.0.10 of this plugin has been initiated at the jenkins update center ... it has the changes from this PR

gabemontero:
PR https://github.com/openshift/jenkins/pull/740 will update the openshift jenkins image with the new version of the plugin, which is not available at the jenkins update center

hrishin:
Thank you!! @gabemontero

gabemontero:
yw @hrishin :-)

andyfeller:
@gabemontero based on the default OpenShift roles to Jenkins permissions, this would be the corresponding `ConfigMap` payload, correct?
```yaml
  data:
    Credentials-Create: admin
    Credentials-Delete: admin
    Credentials-ManageDomains: admin
    Credentials-Update: admin
    Credentials-View: admin,edit,view
    Job-Build: admin,edit
    Job-Cancel: admin,edit
    Job-Configure: admin,edit
    Job-Create: admin,edit
    Job-Delete: admin,edit
    Job-Discover: admin,edit,view
    Job-Read: admin,edit,view
    Job-Workspace: admin,edit
    Overall-Administer: admin
    Overall-Read: admin,edit,view
    Overall-RunScripts: admin,edit
    Run-Delete: admin
    Run-Update: admin
    SCM-Tag: admin,edit
    Slave-Configure: admin
    Slave-Delete: admin
    View-Configure: admin
    View-Create: admin
    View-Delete: admin
```

andyfeller:
@gabemontero: I am testing 1.0.10 release with the following `ConfigMap` and my custom `foobar` OpenShift ClusterRole when I noticed `foobar` not being associated with my matrix auth entry; it only sees the standard view/edit/admin roles as `...-view-edit-admin` is the suffix assigned to me.

```
  data:
    Credentials-Create: admin,foobar
    Credentials-Delete: admin,foobar
    Credentials-ManageDomains: admin,foobar
    Credentials-Update: admin,foobar
    Credentials-View: admin,foobar,edit,view
    Job-Build: admin,foobar,edit
    Job-Cancel: admin,foobar,edit
    Job-Configure: admin,foobar,edit
    Job-Create: admin,foobar,edit
    Job-Delete: admin,foobar,edit
    Job-Discover: admin,foobar,edit,view
    Job-Read: admin,foobar,edit,view
    Job-Workspace: admin,foobar,edit
    Overall-Administer: admin,foobar
    Overall-Read: admin,foobar,edit,view
    Overall-RunScripts: admin,foobar,edit
    Run-Delete: admin,foobar
    Run-Update: admin,foobar
    SCM-Tag: admin,foobar,edit
    Slave-Configure: admin,foobar
    Slave-Delete: admin,foobar
    View-Configure: admin,foobar
    View-Create: admin,foobar
    View-Delete: admin,foobar
```

Looking under the covers, I think my issue has something to do with [`openshift-login` SAR checks against the role list](https://github.com/openshift/jenkins-openshift-login-plugin/blob/master/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java#L608) as it presumes the role is a verb and perhaps something needs to be done with the ClusterRole or binding this cluster role to a user within a project I'm just not aware of.

Appreciate any help!

gabemontero:
Yep the admin or foobar etc. are the verb in the subjectaccessreview that is initiated against the namespace jenkins/login plugin are pointed at.

All part of the RBAC stuff in openshift our security/auth folks support.  Perhaps are readme instructions needs some augmenting.

I would start with suggesting that you read https://docs.okd.io/latest/admin_guide/manage_rbac.html and see if that helps your wire the different rbac piecies correctly.  

If that does not help, we'll need to pull in those guys to help fill the gaps with the subjectaccessreview.

andyfeller:
From https://docs.okd.io/latest/architecture/additional_concepts/authorization.html#evaluating-authorization, verbs for nearly all other resources are:
- `get`
- `list`
- `create`
- `update`
- `delete`
- `deletecollection`
- `watch`
hence the confusion about roles being doubly used as verbs, which I think others will surely trip over, too.  Any thoughts about documenting this aspect?

andyfeller:
So I think I FINALLY understand what is going on here and everything is okay but the way this problem and solution are described is confusing:

1. This PR gives you the ability to map what Jenkins permissions are given to individuals based upon the verbs associated with the OpenShift Roles bound to said individual for the jenkins resource.
2. If you have an issue where project admin is not given to individuals via the `admin` role, then you simply need to add/update the rule associated with whatever non-`admin` role to give the `admin` verb or whatever to those people.  This functionality already existed; just not pointed out in previous conversations.

andyfeller:
This mechanism will be problematic for Jenkins permissions like `Lockable Resources` as ConfigMap data keys are validated against the `[-._a-zA-Z0-9]+` regex.

Issue #46, opened 2018-10-18
make changes to support keycloak


gabemontero:
Change https://github.com/openshift/jenkins-openshift-login-plugin/blob/master/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java#L659-L662 to use https://github.com/openshift/jenkins-openshift-login-plugin/blob/master/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftProviderInfo.java#L52-L55 as the URLs

@enj

gabemontero:
Hey @enj - api.ci busted, I had some cycles to give this a go, but it did not work.

I tried it on a 3.11 cluster brought up via `oc cluster up`.

It should work on 3.11, right?

Assuming so, given this code snippet:

```
         GenericUrl tsu = new GenericUrl(
                getDefaultedServerPrefix() + "/oauth/token");
        String asu = getDefaultedRedirectURL()
                + "/oauth/authorize";
        LOGGER.info("GGM default token endpoint " + tsu);
        LOGGER.info("GGM default authorization endpoint " + asu);
        if (provider != null) {
            LOGGER.info("GGM using provider values");
            tsu = new GenericUrl(provider.token_endpoint);
            asu = provider.authorization_endpoint;
            LOGGER.info("GGM new token endpoint " + tsu);
            LOGGER.info("GGM new authorization endpoint " + asu);
        }
        final GenericUrl tokenServerURL = tsu;
        final String authorizationServerURL = asu;

       final AuthorizationCodeFlow flow = new AuthorizationCodeFlow.Builder(
                BearerToken.queryParameterAccessMethod(), transport,
                JSON_FACTORY, tokenServerURL,
                new ClientParametersAuthentication(getDefaultedClientId(),
                        getDefaultedClientSecret().getPlainText()),
                getDefaultedClientId(), authorizationServerURL).setScopes(
                Arrays.asList(SCOPE_INFO, SCOPE_CHECK_ACCESS)).build();

        final OpenShiftOAuth2SecurityRealm secRealm = this;
        final String url = buildOAuthRedirectUrl(redirectOnFinish);
        LOGGER.info("GGM url " + url);

        return new OAuthSession(flow, from, url) {
            @Override
            public HttpResponse onSuccess(String authorizationCode) {
                try {
                    LOGGER.info("GGM auth code " + authorizationCode);
[...]

PR #49, opened 2018-11-05
Upd oauth urls

gabemontero:
Fixes https://github.com/openshift/jenkins-openshift-login-plugin/issues/46

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1671633

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -393,6 +393,10 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
                 // the hostname or ip/port that is only available within the
                 // cluster
                 this.defaultedRedirectURL = provider.issuer;
+                // for diagnostics: see if the provider endpoints are accessible, given what
+                // Mo told me about them moving the oauth server from internal to a route based external one
+                // on the fly
+                this.useProviderOAuthEndpoint(credential, transport);
             } else {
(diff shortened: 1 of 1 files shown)


gabemontero:
/assign @bparees 

@enj FYI

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -393,6 +393,7 @@ boolean populateDefaults() throws IOException, GeneralSecurityException {
                 // the hostname or ip/port that is only available within the
                 // cluster
                 this.defaultedRedirectURL = provider.issuer;
+                this.testProviderEndpoints(credential, transport);
```
bparees: this doesn't check the return code so what's it doing?

gabemontero: it is a diagnostic,  given what @enj mentioned about the oauth stuff moving around, coming up and down ... I can add a comment

bparees:
nits, otherwise lgtm

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -535,6 +536,55 @@ public void initialize(HttpRequest request)
                 OpenShiftProviderInfo.class);
         return info;
     }
+    
+    private boolean testProviderEndpoints(
```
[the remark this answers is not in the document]
gabemontero: Yep `useProviderOAuthEndpoint` is a better name ... I'll change

PR #51, opened 2018-11-15
dbg / possible fix for roles npe seen by feedhenry img ext

gabemontero:
Fixes https://github.com/openshift/jenkins/issues/743
Fixes https://github.com/openshift/jenkins-openshift-login-plugin/issues/52


gabemontero:
Got a run from @grdryn and the only debug statement was the one in postSAR right before the original NPE happened:

`INFO: DBG postSAR how the heck did roles get set to null?`

I was almost convinced it was the constructor thing ...

I'll do some more cross reference testing with our base image, along with a config map based run, and then productize and merge this, etc.

gabemontero:
v1.0.11 of the login plugin has been initiated

PR #71, opened 2019-05-16
Bug 1709575: account for the jenkins SA certificate differing from th…

gabemontero:
…e oauth server's router certificate (where SSL handshake errors would then occur with jenkins SA certificate) by falling back to the JVM's default keystore

@openshift/sig-developer-experience @jupierce @enj fyi

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -168,6 +168,14 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
      * Will get repopulated on restart
      */
     private static HttpTransport transport;
+    
+    /**
+     * transport that will only leverage the JVMs default keystore and allow for the jenkins SA cert
+     * and the oauth server router cert varying such that SSL handshakes will fail if we exclusively
+     * use the jenkins SA cert
+     * 
+     */
+    private static HttpTransport jvmDefaultKeystoreTransport;
 
     /**
      * The service account directory, if set, instructs the plugin to follow the
@@ -282,6 +290,9 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
         this.serviceAccountName = Util.fixEmpty(serviceAccountName);
 
         OpenShiftOAuth2SecurityRealm.transport = transport;
+        
+        jvmDefaultKeystoreTransport = 
+                new NetHttpTransport.Builder().build();
 
         if (testTransport != null)
             OpenShiftOAuth2SecurityRealm.transport = testTransport;
@@ -396,7 +407,8 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
                 // for diagnostics: see if the provider endpoints are accessible, given what
                 // Mo told me about them moving the oauth server from internal to a route based external one
                 // on the fly
-                this.useProviderOAuthEndpoint(credential, transport);
+                if (this.useProviderOAuthEndpoint(credential))
+                    this.transportToUse(credential);
             } else {
                 runningInOpenShiftPodWithRequiredOAuthFeatures = false;
             }
@@ -540,10 +552,9 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
         return info;
     }
     
-    private boolean useProviderOAuthEndpoint(
-            final Credential credential, final HttpTransport transport) {
+    private HttpTransport transportToUse(final Credential credential) {
         if (provider == null)
-            return false;
+            return transport;
         try {
             HttpRequestFactory requestFactory = transport
                     .createRequestFactory(new HttpRequestInitializer() {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -168,6 +168,13 @@
      * Will get repopulated on restart
      */
     private static HttpTransport transport;
+    
+    /**
+     * transport that will only leverage the JVMs default keystore (allow for the jenkins SA cert
+     * and the oauth server router cert varying such that SSL handshakes will fail
```
adambkaplan: nit - missing close paren

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -574,19 +613,41 @@ public void initialize(HttpRequest request)
...
+            if (rc != 200)
+                LOGGER.info("OpenShift OAuth the attempt to get the server version request got an unexpected return code: " + rc);
+            OpenShiftVersionInfo version = response.parseAs(OpenShiftVersionInfo.class);
+            if (version != null && version.major != null && version.major.equals("1")) {
+                if (version.minor.length() > 2) {
+                    String minor = version.minor.substring(0, 2); // per javadoc end index is not inclusive
+                    int m = Integer.parseInt(minor);
+                    if (m <= 11) {
+                        // 3.x cluster
+                        LOGGER.info("OpenShift OAuth the server is 3.x, specifically " + version.toString());
```
adambkaplan: Are the version numbers for the OpenShift OAuth server completely independent from the product version?

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftVersionInfo.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.openshift.jenkins.plugins.openshiftlogin;
```
adambkaplan: Side note - can't we have a top-level LICENSE file? (if so I'd put this in a separate PR so someone in legal can weigh in)

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftVersionInfo.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2004-2009, Sun Microsystems, Inc., Kohsuke Kawaguchi
```
adambkaplan: I'd strike this, as Sun has nothing to do with this file.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
jupierce: What is being checked here is the underlying kubernetes version. It is effectively independent with 4.0.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: part of next push

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftVersionInfo.java
gabemontero: there are refs to classes constructed at that time  ... I remember having this discussion with Clayton wrt when this plugin first came into being

Not going to diverge at this point

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftVersionInfo.java
gabemontero: devtools can pursue that at this point :-) .... "meets min" wrt effort @adambkaplan :-)

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
gabemontero: Yep it is just a debug marker to note which path wrt transports, oauth endpoints, and certs is being taken

OAuth behavior is different between 3.x and 4.x @adambkaplan , so we vary accordingly.  Goal is to be able to use 1 version of the login plugin across 3.x and 4.x to simplify support.

gabemontero:
comment correction nit pushed @adambkaplan ...explanations/response to the other comments in line with various conversations started

ptal

gabemontero:
I'll be cutting v1.0.17 of the login plugin for this fix

PR #80, opened 2019-09-11
WIP: Fix oauth with ocp 4 and large refactor

akram:
- Adds initialisation of the jvmDefaultKeystoreTransport which was missing
- Turn the jvmDefaultKeystoreTransport into an instance variable. When used as static, it was reusable between users, making the connection conveying the credentials of previously attempted users

- Extracted all anonymous classes into classes
- Reorganised method order: public, getters/setters/protected and then private
- Organized imports and static imports


src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -164,30 +158,30 @@
...
-     * The transport loaded with a service account secret.
-     * Marked static for https://jenkins.io/blog/2018/03/15/jep-200-lts/#after-the-upgrade
-     * Will get repopulated on restart
+     * The transport loaded with a service account secret. Marked static for
+     * https://jenkins.io/blog/2018/03/15/jep-200-lts/#after-the-upgrade Will get
+     * repopulated on restart
      */
-    private static HttpTransport transport;
-    
+    private transient HttpTransport transport;
```
gabemontero: Yeah per you comment `Turn the jvmDefaultKeystoreTransport into an instance variable. When used as static, it was reusable between users, making the connection conveying the credentials of previously attempted users` 

I think I might see how there was a problem, though not necessarily what you have conveyed. 

I would like to walk through some of the details of how you landed there, and want to make sure you went through the same path to this conclusion.

Specifically:
a) there is only one security realm per JVM
b) but given the crappy Jenkins API, we have to recreate that instance each time we add users
c) that said, the transport does not have user specific elements to it .... if you discovered there is, elaborate as to why ... I don't see it
d) but maybe certs rotated and that somehow mess things UP, as the two transport choices needed to be regenerated ?  though they are re-gened in the ctor and populateDefaults

gabemontero:
> After testing it on customer's platform, we have the following error message in the logs:
> 
> ```
> Sep 11, 2019 5:08:23 PM jenkins.security.ClassFilterImpl notifyRejected
> WARNING: com.google.api.client.http.javanet.NetHttpTransport in file:/var/lib/jenkins/plugins/openshift-login/WEB-INF/lib/google-http-client-1.20.0.jar might be dangerous, so rejecting; see https://jenkins.io/redirect/class-filter/
> Sep 11, 2019 5:08:23 PM jenkins.model.Jenkins saveQuietly
> WARNING: null
> java.io.IOException: java.lang.RuntimeException: Failed to serialize jenkins.model.Jenkins#securityRealm for class hudson.model.Hudson
> 	at hudson.XmlFile.write(XmlFile.java:200)
> 	at jenkins.model.Jenkins.save(Jenkins.java:3262)
> 	at jenkins.model.Jenkins.saveQuietly(Jenkins.java:3268)
> 	at jenkins.model.Jenkins.setAuthorizationStrategy(Jenkins.java:2549)
> 	at org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm.updateAuthorizationStrategy(OpenShiftOAuth2SecurityRealm.java:576)
> 	at org.openshift.jenkins.plugins.openshiftlogin.OAuthSessionProcessor.onSuccess(OAuthSessionProcessor.java:42)
> 	at org.openshift.jenkins.plugins.openshiftlogin.OAuthSession.doFinishLogin(OAuthSession.java:129)
> 	at org.openshift.jenkins.plugins.openshiftlogin.OpenShiftOAuth2SecurityRealm.doFinishLogin(OpenShiftOAuth2SecurityRealm.java:647)
> 	at java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:627)
> 	at org.kohsuke.stapler.Function$MethodFunction.invoke(Function.java:396)
> ```
> 
> So, we need also to migrate com.google.oauth-client to a higher version?
> or mark a field as transient in our class so serialisation will be skipped for it.

Undoubtedly it is the latter.  It is the reason any of the fields in that class have been marked transient.  Ran into this error.

You also have to make sure all those transient fields are repopulated in case of jenkins restart.

See for example https://github.com/openshift/jenkins-openshift-login-plugin/blob/master/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java#L318-L334 and the recreation of the permissions filter.

akram:
/hold

@gabemontero The fix was running in a customer environment for a week. But, this PR is getting bigger and bigger. Following your advices and code freeze approaching, I will submit a smaller PR to the fix.

waveywaves:
After the refactor PR it would be better if we rebase and merge new fix commits here. Let's do that ?

akram:
rebased

PR #81, opened 2019-09-16
Bug 1750588: Force initialization of jvmDefaultKeystoreTransport if not initialized

akram:
Fixes: https://github.com/openshift/jenkins/issues/893

On OpenShift Online or OpenShift using an Oauth IDP defined, the jvmDefaultKeystoreTransport is not properly initialised in some cases.

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -24,6 +24,10 @@
  */
 package org.openshift.jenkins.plugins.openshiftlogin;
 
+import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
+import static java.net.HttpURLConnection.HTTP_OK;
+import static java.util.logging.Level.INFO;
+
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
@@ -46,6 +50,7 @@ import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.net.ssl.SSLHandshakeException;
 import javax.servlet.ServletException;
 
 import org.acegisecurity.Authentication;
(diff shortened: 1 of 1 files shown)


akram:
Some changes are re-formatting code

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -532,94 +513,100 @@ public String getLoginUrl() {
...
+            if (hre.getStatusCode() == HTTP_NOT_FOUND) {
+                LOGGER.log(INFO, "OpenShift OAuth got an unexpected 404 trying out the issuer's token endpoint", hre);
             }
-        } catch (javax.net.ssl.SSLHandshakeException ssle) {
-            LOGGER.info("OpenShift OAuth got a SSL error when accessing the issuer's token endpoint when using the SA certificate");
+        } catch (SSLHandshakeException ssle) {
+            String formattedMessage = "OpenShift OAuth got an SSL error when accessing the issuer's token endpoint when using the SA certificate";
+            LOGGER.info(formattedMessage);
             try {
+                if (jvmDefaultKeystoreTransport == null) {
```
gabemontero: cross referencing with the bug's stack trace, there is no NPE at ` HttpRequestFactory requestFactory = jvmDefaultKeystoreTransport
                        .createRequestFactory(new HttpRequestInitializer() {
                        .createRequestFactory(new HttpRequestInitializer() ` a couple of lines below, which we would have gotten with our theory that `jvmDefaultKeystoreTransport` 

But I think that is expected, because of the try/catch/throwable

I don't see a full set of logs in the bugzilla @akram , but if we are correct, we should have seen 

`OpenShift OAuth provider token endpoint failed unexpectedly using the JVMs default keystore` in the log as well as the stack trace shown in the bug, which I think we would have gotten from using the default transport.

Any chance you've
a) seen the entire log and seen the log I mentioned above, or 
b) already validated this patch with the user who saw this when doing the upgrade

I'm guessing the answer is `yes` to at least one of those questions, but let me know.

Pending that clarification, I think we are good here, and we can mark this good to merge.

akram: Hi @gabemontero 

yes, Indeed, I saw the entire stack trace and it was showing the NPE on this value.
And, also, yes, a customer has already tested the fix for about a week.

PR #84, opened 2019-09-19
Bug 1752831: Handle HTTP proxy for HTTPS connections

akram:
Starting Java 7/8 , authenticated http and https proxies did not require to set a custom authenticator if the <protocol>.proxy{Host,Port, User, Password} where set.
However, Basic authentication was disabled and needed to be forcibly re-enabled if needed using 
```
-Djdk.http.auth.tunneling.disabledSchemes=""
-Djdk.http.auth.proxying.disabledSchemes=""
```
This works well for http requests, but https request fails with 
```
java.io.IOException: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 407 Proxy Authentication Required"
```
unless a custom authenticator is set.

This PR sets this authenticator if the https.proxyUser or  https.proxyPassword are set.
Note that the update manager of Jenkins seems to be also impacted by this issue.

Updating google-auth-client has been performed in case it was fixing it, but, no changes. However, I made my tests using it, so I didn't want to revert the change.

/assign @gabemontero @adambkaplan

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -26,6 +26,7 @@ package org.openshift.jenkins.plugins.openshiftlogin;
 
 import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 import static java.net.HttpURLConnection.HTTP_OK;
+import static java.util.logging.Level.FINE;
 import static java.util.logging.Level.INFO;
 import static org.apache.commons.lang.StringUtils.isNotBlank;
 
@@ -36,7 +37,9 @@ import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.net.Authenticator;
 import java.net.MalformedURLException;
+import java.net.PasswordAuthentication;
 import java.net.URL;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
@@ -165,6 +168,10 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
     // NPE issues of this not getting set through based on how this object was
     // constructed
     private static final ArrayList<String> roles = new ArrayList<String>(Arrays.asList("admin", "edit", "view"));
+
+    private static final String HTTPS_PROXY_USER = "https.proxyUser";
+
+    private static final String HTTPS_PROXY_PASSWORD = "https.proxyPassword";
 
     /**
      * Control the redirection URL for this realm. Exposed for testing.
@@ -538,6 +545,7 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm {
     }
 
     private HttpTransport transportToUse(final Credential credential) {
+        initializeHttpsProxyAuthenticator();
         if (provider == null)
             return transport;
         try {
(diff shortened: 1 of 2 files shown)


akram:
Maybe one could ask why here: The OpenShiftOAuth2SecurityRealm makes a connexion to the auth server.
In case of a "disconnected" environment, where nodes/pods cannot access directly the oauth server but needed to go through a proxy, the https query to validate oauth token with the oauth server was failing with:
`java.io.IOException: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 407 Proxy Authentication Required" `

This PR fixes this issue.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -597,6 +605,33 @@ public void initialize(HttpRequest request) throws IOException {
...
+     * If the custom authenticator is not set using (Authenticator.setDefault()) an
+     * error: java.io.IOException: Unable to tunnel through proxy. Proxy returns
+     * "HTTP/1.1 407 Proxy Authentication Required" is thrown.
+     */
+    private void initializeHttpsProxyAuthenticator() {
+        final String username = System.getProperty(HTTPS_PROXY_USER);
+        final String password = System.getProperty(HTTPS_PROXY_PASSWORD);
+        LOGGER.log(INFO, "Checking if HTTPS proxy initialization is required ... ");
+        if (username != null && password != null) {
+            LOGGER.log(INFO, HTTPS_PROXY_USER + " or " + HTTPS_PROXY_PASSWORD + " found in system properties...");
```
ricardomaraschini: Is it advisable to log the password? I guess this is going in clean text, no?

ricardomaraschini: See, my comment is irrelevant.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -597,6 +605,33 @@ public void initialize(HttpRequest request) throws IOException {
...
+            LOGGER.log(INFO, "Now setting a custom PasswordAuthentication authenticator");
```
adambkaplan: nit - first two log statements can probably be set to `DEBUG` - info message will need to be more informative:

```suggestion
            LOGGER.log(INFO, "Creating basic authenticator for HTTPS proxy auth");
```

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -597,6 +605,33 @@ public void initialize(HttpRequest request) throws IOException {
         return transport;
     }
 
+    /**
+     * Since Java 7 and Java 8, it is not required to set a custom Authenticator to
+     * use an authenticated HTTP proxy. However, for HTTPS proxies, this is still
+     * necessary for Basic authentication in conjonction with:
```
adambkaplan: nit - spelling:
```suggestion
     * necessary for Basic authentication in conjunction with:
```

adambkaplan:
/retitle Bug 1752831: Handle HTTP proxy for HTTPS connections

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
akram: Fixed in forced push commit

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
akram: Fixed in forced push commit

gabemontero:
@akram convinced me off line to hold off on the README update, at least for now

/lgtm

PR #91, opened 2019-10-25
Universal redeploy script

akram:


otaviof:
@akram what's this PR about? :-)

akram:
It is a generic plugin redeployment script useful when developing plugins. It allows redeployment of the built plugin in target/ into the remote /var/lib/jenkins/plugins/ of the jenkins instance in the current project :-D

PR #111, opened 2020-09-04
Bug 1813207: Replace INFO with FINE for Oauth,proxy logs

waveywaves:


src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -414,7 +413,7 @@ boolean populateDefaults() throws IOException, GeneralSecurityException {
             if (LOGGER.isLoggable(Level.FINE))
```
akram, line 413: in both cases of the if/else we do the same thing. So, you can remove the if/es

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -553,14 +552,14 @@ private HttpTransport transportToUse(final Credential credential) {
             request.execute().getStatusCode();
         } catch (com.google.api.client.http.HttpResponseException hre) {
             if (hre.getStatusCode() == HTTP_NOT_FOUND) {
-                LOGGER.log(INFO, "OpenShift OAuth got an unexpected 404 trying out the issuer's token endpoint", hre);
+                LOGGER.log(FINE, "OpenShift OAuth got an unexpected 404 trying out the issuer's token endpoint", hre);
```
akram, line 555: keep  this one as INFO, we need to see it if this happens.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -553,14 +552,14 @@ private HttpTransport transportToUse(final Credential credential) {
...
             }
         } catch (SSLHandshakeException ssle) {
             String formattedMessage = "OpenShift OAuth got an SSL error when accessing the issuer's token endpoint when using the SA certificate";
             LOGGER.info(formattedMessage);
             try {
                 if (jvmDefaultKeystoreTransport == null) {
-                    LOGGER.log(INFO, "jvmDefaultKeystoreTransport was not initialized: Forcing initalization");
+                    LOGGER.log(FINE, "jvmDefaultKeystoreTransport was not initialized: Forcing initalization");
```
akram, line 562: keep this one also as INFO, it is poor likely to happen, so if it happens, it means there is a problem, we need to see it

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -572,22 +571,22 @@ private HttpTransport transportToUse(final Credential credential) {
...
             } catch (com.google.api.client.http.HttpResponseException hre) {
                 // this means SSL handshakes work, but our generic head simply is not honored by
                 // the endpoint
-                LOGGER.info(
+                LOGGER.fine(
                         "OpenShift OAuth was able to complete the SSL handshake when accessing the issuer's token endpoint using the JVMs default keystore");
                 return jvmDefaultKeystoreTransport;
             } catch (Throwable t) {
-                LOGGER.log(Level.INFO,
+                LOGGER.log(Level.FINE,
```
akram, line 583: keep this one as INFO, it is in an try/catch block.

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -572,22 +571,22 @@ private HttpTransport transportToUse(final Credential credential) {
...
                         "OpenShift OAuth provider token endpoint failed unexpectedly using the JVMs default keystore",
                         t);
                 return jvmDefaultKeystoreTransport;
             }
         } catch (Throwable t) {
-            LOGGER.log(Level.INFO,
+            LOGGER.log(Level.FINE,
```
akram, line 589: same thing here, must stay as INFO, even error would be ok

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -875,7 +874,7 @@ protected OAuthSession newOAuthSession(String from, final String redirectOnFinis
                 transportForThisRequest));
         LOGGER.info(format("AuthorizationCodeFlow using : clientAuthentication=[%s], clientId=[%s], ",
                 clientAuthentication, defaultedClientId));
-        LOGGER.info(format("AuthorizationCodeFlow using : authorizationServerURL=[%s], ", authorizationServerURL));
+        LOGGER.info(format("AuthorizationCodeFlow using : authorizationSeqrverURL=[%s], ", authorizationServerURL));
```
akram, line 877: keep it as info, and remove the added typo

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -1211,7 +1210,7 @@ protected void deleteOauthAccessToken(String oAuthAccessToken) {
             }
             LOGGER.info("oAuthAccessToken successfully deleted");
         } catch (Throwable t) {
-            LOGGER.log(Level.INFO, "Failed to delete oAuthAccessToken", t);
+            LOGGER.log(Level.FINE, "Failed to delete oAuthAccessToken", t);
```
akram, line 1213: keep it in INFO as it it is in try/catch

akram (changes requested):

PR #113, opened 2020-09-15
Bug 1879327: The OCP version <= 4.5,the oAuthtokenName == oAuthToken

barleyer:
@akram when OCP version <= 4.5,the `oAuthtokenName` should be same with `oAuthToken`.

(The token in OCP >= 4.6,should start with sha256~)

### DIFF
--- a/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
+++ b/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
@@ -195,7 +195,7 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm implements Seria
      * transport that will only leverage the JVMs default keystore and allow for the
      * jenkins SA cert and the oauth server router cert varying such that SSL
      * handshakes will fail if we exclusively use the jenkins SA cert
-     * 
+     *
      */
     private static HttpTransport jvmDefaultKeystoreTransport;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -1154,7 +1154,10 @@ public void doLogout(StaplerRequest req, StaplerResponse resp) throws IOExceptio
         if (idTokenResponse != null) {
             LOGGER.info("Found an  oauthaccess token in the sessions");
             String oAuthToken = idTokenResponse.getAccessToken();
-            String oAuthtokenName = tokenToObjectName(oAuthToken);
+            String oAuthtokenName = oAuthToken;
```
akram, line 1157: Hi @barleyer , thanks for reporting this and submitting the PR. 
regarding this code portion, it is preferable to compute the `oAuthtokenName` consistently in the  `tokenToObjectName ` method and thus to add this test in this method.

based on your tests, and your comments, can you confirm that for 4.5- ,  idTokenResponse.getAccessToken() is the actual token name?

On our side, we haven't released the 4.5 image that tests this.

akram (changes requested):

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
barleyer: Yeah,the value of token is the object name of the token.

akram:
as discussed on slack, I am closing this one in favor of #114 
we will use also the bz for the release of the jenkins image for 4.5 , not for the fix in the plugin

cc @jkhelil

PR #122, opened 2021-10-27
Update README.md - add more detailed instructions with how to use custom role/verb beyond just `view`,`edit`,`admin`

itewk:
The current read me implies that all you need to do to use custom "role"s beyond `admin`, `edit`, and `view` is to create a `Role` and bind users/groups to it. That is not actaully the case. after hours of fighting and digging through code figured out the magic that the plugin isn't actauly looking at Roles it is looking for access to verbs on the `build.openshift.io` groupAPI `jenkins` resource. Which makes sense with the way SubjectAccessRequests work, but really needs to be better documented for the next soul to try and do this.

This is my attempt at clearer documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,66 @@ time the poll occurs.
 
 You can control how often the polling occurs with the `OPENSHIFT_PERMISSIONS_POLL_INTERVAL` environment variable.  The default polling interval when no environment variable is set is 5 minutes.
 
+#### Using custom Roles
+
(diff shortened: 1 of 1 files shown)


itewk:
@otaviof @sbose78 thoughts?

itewk:
ping?

itewk:
ping

akram:
hi @itewk , thank you for reverse engineering this and for the documentation update.
I will merge it now.

itewk:
@akram wahooo. thank you!

PR #145, opened 2023-05-03
Make name of permission mapping ConfigMap configurable

### LINKED ISSUE #144: Two instances in the same namespace use the same ConfigMap for permission mapping
The OpenShiftOAuth2SecurityRealm has the name of the ConfigMap for permission mapping hard-coded into the CONFIG_MAP_URI field ("openshift-jenkins-login-plugin-config"):

https://github.com/openshift/jenkins-openshift-login-plugin/blob/master/src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java#L152

`private static final String CONFIG_MAP_URI = "/api/v1/namespaces/%s/configmaps/openshift-jenkins-login-plugin-config";`

We would like to set the name of the ConfigMap via environment variables in order to have more than one Jenkins instance in the same namespace with different permission mapping ConfigMaps.

stef-h:
resolves #144

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,8 +98,7 @@ Now, with v1.0.10 of this plugin, you can change
 * Which OpenShift Roles are checked for authorization
 * Which Jenkins permissions map to which OpenShift Roles
 
-This plugin will look at for a ConfigMap named openshift-jenkins-login-plugin-config.  Typically, when running Jenkins in an OpenShift Pod, it 
-will look in the namespace that Jenkins is running in.  Otherwise, it looks in the namespace specified in the "client ID" as explained in ["Secondary Scenarios" down below.](#secondary-scenarios) 
+This plugin will look for a ConfigMap named after the content of the environment variable `CONFIG_MAP_NAME`.  The default ConfigMap name when no environment variable is set is "openshift-jenkins-login-plugin-config".  Typically, when running Jenkins in an OpenShift Pod, it will look in the namespace that Jenkins is running in.  Otherwise, it looks in the namespace specified in the "client ID" as explained in ["Secondary Scenarios" down below.](#secondary-scenarios) 
 
 If this plugin finds and can read in that ConfigMap, it then:
 
--- a/README.md
+++ b/README.md
@@ -98,7 +98,9 @@ Now, with v1.0.10 of this plugin, you can change
 * Which OpenShift Roles are checked for authorization
 * Which Jenkins permissions map to which OpenShift Roles
 
-This plugin will look for a ConfigMap named after the content of the environment variable `CONFIG_MAP_NAME`.  The default ConfigMap name when no environment variable is set is "openshift-jenkins-login-plugin-config".  Typically, when running Jenkins in an OpenShift Pod, it will look in the namespace that Jenkins is running in.  Otherwise, it looks in the namespace specified in the "client ID" as explained in ["Secondary Scenarios" down below.](#secondary-scenarios) 
+By default, this plugin looks for a ConfigMap named "openshift-jenkins-login-plugin-config" in the namespace that Jenkins is running in. You can override this behavior by specifying an environment variable named `CONFIG_MAP_NAME` that points to a well formatted ConfigMap that contains your custom configuration.
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -149,7 +149,8 @@ public class OpenShiftOAuth2SecurityRealm extends SecurityRealm implements Seria
 
     private static final String USER_URI = "/apis/user.openshift.io/v1/users/~";
     private static final String SAR_URI = "/apis/authorization.openshift.io/v1/subjectaccessreviews";
-    private static final String CONFIG_MAP_URI = "/api/v1/namespaces/%s/configmaps/openshift-jenkins-login-plugin-config";
+    private static final String CONFIG_MAP_DEFAULT_NAME = "openshift-jenkins-login-plugin-config";
```
coreydaley, line 152:
```suggestion
    private static final String CONFIG_MAP_NAME_DEFAULT = "openshift-jenkins-login-plugin-config";
```

src/main/java/org/openshift/jenkins/plugins/openshiftlogin/OpenShiftOAuth2SecurityRealm.java
```diff
@@ -448,6 +454,12 @@ boolean populateDefaults() throws IOException, GeneralSecurityException {
                     this.serverPrefix));
             LOGGER.info(format(msg5, this.defaultedServerPrefix));
         }
+
+        configMapName = EnvVars.masterEnvVars.get("CONFIG_MAP_NAME");
+        if (configMapName == null) {
+            configMapName = CONFIG_MAP_DEFAULT_NAME;
+        }
```
coreydaley, line 461: I haven't tested the below code, but I think it conveys that I think we should set the configmapName to the default first, and then if the env var is set, change it to that value.
```suggestion
        configMapName = CONFIG_MAP_NAME_DEFAULT;
        if (EnvVars.masterEnvVars.get("CONFIG_MAP_NAME") != null) {
            configMapName = EnvVars.masterEnvVars.get("CONFIG_MAP_NAME");
        }
```

README.md
```diff
@@ -98,8 +98,7 @@ Now, with v1.0.10 of this plugin, you can change
 * Which OpenShift Roles are checked for authorization
 * Which Jenkins permissions map to which OpenShift Roles
 
-This plugin will look at for a ConfigMap named openshift-jenkins-login-plugin-config.  Typically, when running Jenkins in an OpenShift Pod, it 
-will look in the namespace that Jenkins is running in.  Otherwise, it looks in the namespace specified in the "client ID" as explained in ["Secondary Scenarios" down below.](#secondary-scenarios) 
+This plugin will look for a ConfigMap named after the content of the environment variable `CONFIG_MAP_NAME`.  The default ConfigMap name when no environment variable is set is "openshift-jenkins-login-plugin-config".  Typically, when running Jenkins in an OpenShift Pod, it will look in the namespace that Jenkins is running in.  Otherwise, it looks in the namespace specified in the "client ID" as explained in ["Secondary Scenarios" down below.](#secondary-scenarios)
```
coreydaley, line 101: By default, this plugin looks for a ConfigMap named "openshift-jenkins-login-plugin-config" in the namespace that Jenkins is running in. You can override this behavior by specifying an environment variable named `CONFIG_MAP_NAME` that points to a well formatted ConfigMap that contains your custom configuration.  

Typically, when running Jenkins in an OpenShift Pod, it will look in the namespace that Jenkins is running in.  Otherwise, it looks in the namespace specified in the "client ID" as explained in ["Secondary Scenarios" down below.](#secondary-scenarios)

coreydaley (changes requested):
Once you have made these changes I will run the tests and I think we will be good to go!

Thank you for your contribution!

stef-h:
> Once you have made these changes I will run the tests and I think we will be good to go!
> 
> Thank you for your contribution!

Thanks for the suggested improvements. I added them to the PR

stef-h:
Hi @coreydaley , from the output of the two failing tests I cannot really tell what went wrong and if it is related to my changes in any way. Can you please tell me how I can support in resolving the problems with the two failing test cases?

PR #212, opened 2024-08-28
updating builder image to use JDK21 based image

ayushsatyam146:
Updating builder image to UBI9 JDK21 Image and removing the references to install maven externally


sayan-biswas:
We can use the `1.20` tag I guess, instead of using a digest.

sayan-biswas (approved):

